PR #5, opened 2024-01-25
Blueprint of elementary consequences

teorth:
Added a (slightly sketchy) blueprint of the various elementary consequences of the PNT as another file.  (Someone will need to put in \leans and \uses to flesh it out.)


AlexKontorovich (approved):
Thanks!

PR #9, opened 2024-01-30
fix: remove unneeded lean_exe target to improve build time

dwrensha:
Removes the `lean_exe` build target, which was making `lake build` take a long time because it was unable to use the mathlib cache. Makes `lean_lib PrimeNumberTheoremAnd` the default build target. After this change, the mathlib cache does get used and things should go much faster.

### DIFF
--- a/Main.lean
+++ b/Main.lean
@@ -1,1 +1,0 @@
-import «PrimeNumberTheoremAnd»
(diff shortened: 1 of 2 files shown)


ianjauslin-rutgers (approved):
Awesome, thanks! I was wondering why the cache wasn't being used...

PR #10, opened 2024-01-30
Blueprint fixes

ianjauslin-rutgers:
Alright, I have things set up so that the blueprint gets generated first, then the doc. While the doc is being generated, the new blueprint is uploaded, along with the previous version of the doc (which may be out of date until the new doc is generated, but at least there is something).


AlexKontorovich (approved):
Great thanks!

PR #13, opened 2024-01-31
Spellings & leanoks for consequences.lean

teorth:
Added spellings of lemmas for consequences.lean and added leanok's.  Also fixed one bad \uses in Wiener.lean.


AlexKontorovich (approved):
Thanks!

PR #14, opened 2024-01-31
Add some helper lemmas comparing different types of sums to Consequen…

teorth:
Some simple lemmas to relate truncated sums over reals to truncated sums over natural numbers, in anticipation of this being useful in some of the proofs in consequences.lean


AlexKontorovich (approved):
Nice, thanks!

PR #16, opened 2024-01-31
Define and prove SmoothExistence lemma

BoltonBailey:
Adds the lemma. Also modifies the Urysohn argument to make it a bit easier.


BoltonBailey:
Depends on #15

AlexKontorovich (approved):
Thank you!!

PR #17, opened 2024-02-01
fix: stale docs for PrimeNumberTheoremAnd.*

llllvvuu:
Currently the docs don't reflect lemmas being moved from `AuxiliaryLemmata` to `MellinCalculus`. This is because `.lake/build/doc/declarations/declaration-data-PrimeNumberTheoremAnd*` are being cached.

This is the same fix as https://github.com/teorth/pfr/pull/173.


AlexKontorovich (approved):
Thanks for fixing this!

PR #20, opened 2024-02-01
Fix nonegativity in SmoothExistence

BoltonBailey:
I'm noticing that the LaTeX stipluates this function should be nonnegative, so I've fixed that


AlexKontorovich (approved):
Thanks!

PR #27, opened 2024-02-03
feat: `tendsto_Realpow_atTop_nhds_0_of_norm_gt_1`

modderme123:
This felt kinda awkward to express... I wonder if there is a tactic that could make applying `Tendsto.comp` more direct?

Update: using `convert` and `simp only` simplified the proof quite a bit. New proof is from @Ruben-VandeVelde


teorth:
> This felt kinda awkward to express... I wonder if there is a tactic that could make applying `Tendsto.comp` more direct?

Maybe you can open a thread about this on the PNT Zulip?  There are often many good suggestions from the community there.

AlexKontorovich (approved):
Thanks!

PR #32, opened 2024-02-05
Implement pdf generation

ianjauslin-rutgers:
I implemented pdf generation. To do so, `cd` to `blueprint` and run `make pdf`.


AlexKontorovich (approved):
Thanks!!

PR #28, opened 2024-02-04
feat: `isIntegrable`, `isHolomorphicOn`

llllvvuu:
I also introduce a `PrimeNumberTheoremAnd/Mathlib` folder for organizing potential mathlib contributions.

PR #35, opened 2024-02-07
ci: check PRs

llllvvuu:


modderme123:
Looks like the build failures are because #33 and #28 had conflicts (deleting the local copy of `integrable_iff_integrableAtFilter_cocompact` in favor of your mathlib version should fix that, do you want me to do that in a separate PR?)

PR #36, opened 2024-02-07
Test

ianjauslin-rutgers:
A test that breaks build


modderme123:
To cause the build to fail, I think you'd want to import Basic.lean here https://github.com/AlexKontorovich/PrimeNumberTheoremAnd/blob/main/PrimeNumberTheoremAnd.lean

PR #40, opened 2024-02-08
partial proof for DiffVertRect_eq_UpperLowerUs

mo271:


mo271:
This is not yet finished, but I don't have time to continue during the next days, so I though I make a pr with what I have...

AlexKontorovich (approved):
Thanks!

PR #44, opened 2024-02-13
bump

Ruben-VandeVelde:


mo271:
Meanwhile there is an even newer version of of EulerProducts, so we might bump to that: https://github.com/MichaelStollBayreuth/EulerProducts/pull/1

AlexKontorovich (approved):
Thanks!

PR #45, opened 2024-02-14
fix LowerUIntegral

mo271:


mo271:
I assume this is what was meant.

mo271:
The remaining two sorrys will be easy with https://github.com/leanprover-community/mathlib4/pull/10485, for which we need to bump mathlib.

PR #57, opened 2024-02-17
refactor: golf using RectangleBorderIntegrable API

llllvvuu:
imports and golfs some lemmas from https://github.com/leanprover-community/mathlib4/pull/9598


AlexKontorovich (approved):
Thanks!!

PR #62, opened 2024-02-18
refactor: generalize to `RectanglePullToNhdOfPole'`

llllvvuu:
uses the new `Square` API and adds some more square and rectangle lemmas


AlexKontorovich (approved):
Thanks!

PR #73, opened 2024-02-20
Move rectangle API to a separate file, and add lemmas about rectangle integrals of `1 / (s - p)`

vbeffara:


AlexKontorovich (approved):
Thanks!

PR #77, opened 2024-02-22
refactor: generalize sigma pulls

llllvvuu:
I ended up still using `RectanglePullToNhdOfPole'` in `verticalIntegral_sub_verticalIntegral_eq_squareIntegral` since contour pulling directly to a square would use more integrability assumptions.

This proof also skips all of the U integral work (`DiffVertRect_eq_UpperLowerUs`, `RectangleIntegral_tendsTo_UpperU`, `HolomorphicOn.upperUIntegral_eq_zero`, etc), but since those lemmas are of general use I kept them.


AlexKontorovich (approved):
Thanks!

PR #79, opened 2024-02-22
More detail on decay bound with explicit const

teorth:
Added more detail on how to prove decay estimate with an explicit constant


AlexKontorovich (approved):
Thanks!

PR #98, opened 2024-03-02
Blueprint proofs

ianjauslin-rutgers:
I wrote some more detail for a couple of proofs in the blueprint. More to come.

I think I did the proof of `MellinOfDeltaSpikeAt1_asymp` in a different way from what you originally intended (I'm not using differentiability). Let me know if it works for you.


AlexKontorovich (approved):
Thanks!

PR #99, opened 2024-03-02
Blueprint proofs

ianjauslin-rutgers:
A couple more proofs.

I added an assumption in `Smooth1Properties_below` that x>0 and in `Smooth1Properties_above` that epsilon is <1. I also split off a technical estimate into its own lemma.


AlexKontorovich (approved):
Thanks!

PR #105, opened 2024-03-04
Blueprint proofs

ianjauslin-rutgers:
More proofs

I also added a lemma that f*g=g*f, which is useful to prove an integrability result for MellinOfSmooth1a.


PrimeNumberTheoremAnd/MellinCalculus.lean
```diff
@@ -935,13 +975,54 @@ lemma MellinOfSmooth1a (Ψ : ℝ → ℝ)
...
+  =\frac1s\mathcal M(\psi)(\epsilon s)
+  .
+$$
 \end{proof}
 %%-/
+
 /-%%
 \begin{lemma}[MellinOfSmooth1b]\label{MellinOfSmooth1b}\lean{MellinOfSmooth1b}\leanok
-For any $s$, we have the bound
+Given $0<\sigma_1\leqslant\sigma_2$, for any $s$ such yjay $\sigma_1\le\mathcal Re(s)\le\sigma_2$, we have
```
AlexKontorovich:
```suggestion
Given $0<\sigma_1\le\sigma_2$, for any $s$ such that $\sigma_1\le\mathcal Re(s)\le\sigma_2$, we have
```

PR #112, opened 2024-03-07
Proposition 3 (moebius bound)

grhkm21:
Prove https://alexkontorovich.github.io/PrimeNumberTheoremAnd/web/sect0005.html#mun Proposition 3.

The ✓ is not appearing next to the theorem, and I don't know how to add it.

I will also try Theorem 22 and Corollary 5 (in the same file).


ianjauslin-rutgers (commented):
Looks great, thank you!

AlexKontorovich (approved):
Thanks!

PR #113, opened 2024-03-08
Added proof of convergence of mu(n)/n to zero.

teorth:
Following Zulip discussion, added to the blueprint a proof that mu(n)/n sums conditionally to zero.


AlexKontorovich (approved):
Thanks!

PR #114, opened 2024-03-08
Fix in blueprint proof: \int |fg| \le \int (\sup |f|) |g| only if |g| is integrable

ianjauslin-rutgers:
I noticed an issue with one of the proofs I wrote: I used the fact that the integrands in an integral are non-negative and bounded to bound an integral. However, because the integral of non_integrable functions is 0, this is only true under an integrability condition. I added this to the proof.


AlexKontorovich (approved):
Thanks!

PR #115, opened 2024-03-09
Fix README blueprint instructions

grhkm21:
Sorry for the mistake, I introduced it in my previous PR.


AlexKontorovich (changes requested):
Can you please also change "irst install" just above to "first install"?

AlexKontorovich (approved):
Thanks!

PR #132, opened 2024-03-29
Lift from `Int` to `Nat` in a cleaner way

FLDutchmann:
The `lift` tactic didn't work before because `n` was assigned a value by the `let` binding, which made it impossible to do the substitution. I made it work by creating an equational hypothesis `hn` using the `set` tactic and then clearing the value of `n`.


AlexKontorovich (approved):
Great thanks!

PR #151, opened 2024-04-08
Wrote a quick human proof of PNT in AP

teorth:
Added a blueprint for how to use Wiener-Ikehara to prove the PNT in AP, but did not provide a formalization.


AlexKontorovich (approved):
Thanks!

PR #161, opened 2024-04-17
Big refactor

vbeffara:
This separates out some API about Sobolev-like spaces and compactly supported functions, to make proofs more readable.


AlexKontorovich (approved):
Thanks!

PR #162, opened 2024-04-18
feat: `ZetaNear1BndExact`, `Real.differentiableAt_cpow_const_of_ne`, `Finset.sum_Ioc_add_sum_Ioc`, `tendsto_coe_atTop`, progress with `finsetSum_tendsto_tsum`, `isPathConnected_aux`, `ZetaUpperBnd`, `ZetaDerivUpperBnd` and `SmoothedChebyshevDirichlet`

VladaSedlacek:

PR #163, opened 2024-04-20
refactor: golf and style improvements

VladaSedlacek:
This also includes the progress from #162 (and hence should be merged after it).


AlexKontorovich (approved):
Thanks!

PR #168, opened 2024-04-29
Fix blueprint

ianjauslin-rutgers:
I fixed the blueprint. There was a '$' missing (and also a '\end{proof}').

I also fixed the versions of the dependencies. This has nothing to do with the problem, but it may avoid issues in the future.


AlexKontorovich (approved):
Thanks!!

PR #178, opened 2024-05-29
bump mathlib

sgouezel:
This makes it possible to delete two files, and remove a lot of things in the file `Fourier.lean`, taking advantage of the fact that mathlib now knows that Fourier transform is a bijection on Schwartz space.

(You might need to delete your `.lake` subdirectory to get things working again, as there have been pretty dramatic changes on the `lake` side).


AlexKontorovich (approved):
Thank you!!

PR #180, opened 2024-05-30
Refactor a few proofs

pitmonticone:
- [x] Substitute `simp` with `simp only` where is appropriate (see [this Zulip topic](https://leanprover.zulipchat.com/#narrow/stream/287929-mathlib4/topic/flexible.20vs.20rigid.20tactics) for more details)
- [x] Use proof terms where possible


AlexKontorovich (approved):
Thanks!

PR #183, opened 2024-06-25
Revert mathlib version to a fixed commit

ianjauslin-rutgers:
I fixed the version of mathlib to a commit from before anything broke. Going forward, when someone needs to upgrade mathlib, they should upgrade to a fixed version of mathlib, not 'master'


AlexKontorovich (approved):
Thank you!!

PR #184, opened 2024-08-23
Chebyshev_asymptotic

teorth:


teorth:
This is a proof of the Chebyshev form $\sum_{p \leq x} \log p ~ x$ of the prime number theorem from the von Mangoldt form.  (Was surprisingly tricky, but hopefully some of the lemmas here can be reused.)

AlexKontorovich (approved):
Thanks!

PR #188, opened 2024-09-14
chore: drop local IntegrableAtFilter.sup_iff

Ruben-VandeVelde:
This was upstreamed in the correct MeasureTheory namespace


AlexKontorovich (approved):
Thanks!

PR #194, opened 2024-10-15
Bump doc-gen4 to v4.12.0

collares:
`make doc` succeeds. cc @AlexKontorovich

### DIFF
--- a/lakefile.lean
+++ b/lakefile.lean
@@ -24,4 +24,4 @@ require EulerProducts from git
   "https://github.com/MichaelStollBayreuth/EulerProducts.git" @ "024a986"
 
 meta if get_config? env = some "dev" then require «doc-gen4» from git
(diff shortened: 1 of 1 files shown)


pitmonticone:
Oh, didn't noticed this before opening #195. 

It seems you've not included the relevant dependencies in manifest though.

collares:
> It seems you've not included the relevant dependencies in manifest though.

I think the workflow here is that `make doc` runs `lake update` to add it in CI: https://github.com/AlexKontorovich/PrimeNumberTheoremAnd/blob/main/Makefile#L14-L17. `doc-gen4` hasn't been part of the committed manifest since March 13th (3bd7dfbd3878fb860a52404af469b1b434bf111d).

pitmonticone:
Oh, I see. Thank you for pointing that out.

PR #197, opened 2024-10-17
add blueprint proof of WeakPNT-character

teorth:
Expand on the proof of the first component, WeakPNT-character, and fix a typo


AlexKontorovich (approved):
Thanks!

PR #202, opened 2024-11-02
added consequences of Dirichlet

teorth:
Added some standard consequences of the PNT in APs, most notably Dirichlet's theorem that primitive residue classes contain infinitely many primes.


AlexKontorovich (approved):
Thanks!

PR #211, opened 2024-11-28
`pi_asymp` and `pi_alt`

jjaassoonn:
Co-authored-by: Junqi Liu <@ahhwuhu>


teorth (approved):
Wow, that was quite an argument!  Is there anything you can think of that might have allowed for a simpler route to get these conclusions?  It's remarkable how hard it is to manipulate these sorts of estimates involving O() and o() bounds even though it is "just" using the triangle inequality and summation by parts...

PR #213, opened 2024-12-09
Prove integrable_x_mul_Smooth1 

FLDutchmann:
Proving measurability was much more painful than expected. I hope that can be golfed down quite a bit. Then again, maybe the right thing to do here is to prove the relevant smoothness properties of the mellin convolution.

I also proved a lemma `support_MellinConvolution_subsets ` that says if $f, g$ are supported on $A, B$ respectively then their mellin convolution is supported on $A*B$. Annoyingly this wasn't actually useful here, since `DeltaSpike` is not generally supported on $[0, \infty)$. Still, I think this is useful API to have.


AlexKontorovich (approved):
Thanks!

PR #214, opened 2025-01-04
Bump to Mathlib 4.14

ajirving:
Bump to Mathlib version 4.14 and the latest commit of EulerProducts.  Summary of changes:
- The notation for infinitely differentiable functions has changed, see https://github.com/leanprover-community/mathlib4/pull/19442.
- Some results from Rectangle.lean are now in Mathlib.  It's possible that more results in that file are now either in Mathlib or unused but I wasn't sure about removing them.
- Various results from EulerProducts are now in Mathlib.  IN particular the derivation of WeakPNT from Wiener uses results purely in Mathlib.  It mostly specializes results for arithmetic progressions so it should be easy to extend to the AP case.  We currently still depend on some small results from EulerProducts.Auxiliary which we should maybe consider copying so we can remove the dependency.
- Updates for a number of theorems which were renamed in Mathlib.

Possible future work since this PR is already quite large:
- Use Abel summation from Mathlib in Consequences.lean rather than the version proved there.
- Remove all dependence on EulerProducts.
- Attempt to improve the compile times of the larger files, especially Wiener.lean and Consequences.lean, as they're painful to work with.


PrimeNumberTheoremAnd/ZetaBounds.lean
```diff
@@ -1844,7 +1843,9 @@ and continuity on a compact interval otherwise.
 $|\zeta(x)^3 \cdot \zeta(x+iy)^4 \cdot \zeta(x+2iy)| \ge 1$. -/
 lemma norm_zeta_product_ge_one {x : ℝ} (hx : 0 < x) (y : ℝ) :
     ‖ζ (1 + x) ^ 3 * ζ (1 + x + I * y) ^ 4 * ζ (1 + x + 2 * I * y)‖ ≥ 1 := by
-  have ⟨h₀, h₁, h₂⟩ := one_lt_re_of_pos y hx
+  have h₀ : 1 < ( 1 + x : ℂ).re := by simp[hx]
+  have h₁ :  1 < (1 + x + I * y).re  := by simp[hx]
+  have h₂ :  1 < (1 + x + 2 * I * y).re := by simp[hx]
```
Ruben-VandeVelde:
```suggestion
  have h₁ : 1 < (1 + x + I * y).re := by simp [hx]
  have h₂ : 1 < (1 + x + 2 * I * y).re := by simp [hx]
```

lake-manifest.json
```diff
@@ -55,91 +45,61 @@
...
-   "name": "BibtexQuery",
+   "scope": "leanprover",
+   "rev": "726b3c9ad13acca724d4651f14afc4804a7b0e4d",
+   "name": "Cli",
    "manifestFile": "lake-manifest.json",
-   "inputRev": "master",
+   "inputRev": "main",
    "inherited": true,
-   "configFile": "lakefile.lean"},
-  {"url": "https://github.com/leanprover/doc-gen4.git",
```
Ruben-VandeVelde, line 134: I think this means you `lake update`d incorrectly (not unexpectedly, I always get caught out as well)

PrimeNumberTheoremAnd/Rectangle.lean
```diff
@@ -11,14 +11,6 @@ variable {z w : ℂ} {c : ℝ}
 This files gathers definitions and basic properties about rectangles.
 %%-/
 
-/-%%
-\begin{definition}\label{Rectangle}\lean{Rectangle}\leanok
-A Rectangle has corners $z$ and $w \in \C$.
-\end{definition}
-%%-/
```
Ruben-VandeVelde, line 18: Does this impact the blueprint? How should we deal with this?

Ruben-VandeVelde (commented):
Looks great!

PrimeNumberTheoremAnd/Rectangle.lean
ajirving: Good point.  I think there's only one place which references the Rectangle definition so I can either remove Rectangle from that \uses line or I could put back the LaTeX definition so the blueprint is unchanged.

lake-manifest.json
ajirving: I just updated the revisions in lakefile.lean and then ran lake update. What did I do wrong and what's the fix?

AlexKontorovich:
@ianjauslin-rutgers could you please check that there won't be a problem with the new lake-manifest.json file and our custom blueprint setup? Thanks!

lake-manifest.json
collares: You still need to update the doc-gen4 commit in the lakefile. Note that it's OK to remove doc-gen from `lake-manifest.json` (or not have it in the first place) because it gets added during CI. If you do have it, it shouldn't cause trouble, as CI will just ignore/overwrite it.

PrimeNumberTheoremAnd/Rectangle.lean
ianjauslin-rutgers: Great! Since the Rectangle definition is no longer used, could you please remove the `Rectangle` from the `\uses` in `GeneralMeromorphic.lean`, `MediumPNT.lean`? Thank you!

ajirving: Done

lake-manifest.json
ajirving: I updated to v4.14.0 of doc-gen4 and reran lake update.  The .json is now the same number of lines as on main but is significantly reordered so hard to compare, does it look OK?

ianjauslin-rutgers (approved):
Awesome, thank you!

PR #215, opened 2025-01-09
Abel summation from Mathlib

ajirving:
Remove abel_summation from Consequences.lean as there's now a version in Mathlib we can use.  The Mathlib version is almost identical with just some minor reordering and notation changes.


AlexKontorovich (approved):
Thanks!

PR #221, opened 2025-02-01
Move limit out of consequences.lean

ajirving:
Moves Real.tendsto_pow_log_div_pow_atTop out of Consequences.lean into a file under the Mathlib directory.  The main reason is that I need this limit for estimating the derivative in ZetaBounds.lean.  It also seems like a good candidate for an easy PR to Mathlib.  I chose the location to match that of tendsto_pow_log_div_mul_add_atTop, which this theorem depends on.


AlexKontorovich (approved):
Sounds good, thanks!

PR #223, opened 2025-02-02
DerivUpperBnd_aux7 

ajirving:
Completes the proof of DerivUpperBnd_aux7.  As expected I needed to change the constant; I updated the rest of the file accordingly and it now builds without warnings.

I didn't want to make any more changes to the statement of the theorem but I think it could be cleaned up.  For example the assumptions $0 < N \leq |t|$ can be inferred from the definition of N and that $|t| > 3$.  I also can't see where $s \ne 1$ is being used but Lean isn't warning about an unused assumption so maybe I'm missing something.  It probably also only needs that $\sigma > 1/2$.

I found that the integral could be estimated and evaluated directly on the infinite interval rather than using limits of finite integrals.  Maybe this approach could be used to slightly simplify the proof of ZetaBnd.


AlexKontorovich (approved):
Thanks!

PR #231, opened 2025-02-16
Update blueprint forSmoothedChebyshevDirichlet_aux_integrable

ajirving:
Fixes a typo in the statement of the theorem and also updates the proof. I think the previous proof didn't work because $x^{1+it}$ isn't integrable, what we need is the quadratic decay of the Mellin transform. It might be tidier not to expand the definition of Mellin transform in this statement but that can wait until someone formalizes the proof.


AlexKontorovich (approved):
Thanks!

PR #236, opened 2025-02-22
Smooth1 properties

ajirving:
Quite a lot here:
* Proves SmoothedChebyshevDirichlet_aux_contAt which I have renamed to Smooth1ContinuousAt and moved to MellinCalculus.lean.
* Given continuity of Smooth1 and the existing bounds on it we can deduce that it is Mellin convergent and it's Mellin transform is differentiable in $\Re s > 0$, see Smooth1MellinConvergent and Smooth1MellinDifferentiable. The latter currently isn't used but will be needed when we want to pull contours, it also gives us continuity of the Mellin transform which will be useful for (SmoothedChebyshevDirichlet_aux_integrable.
* (integrable_x_mul_Smooth1 is now not needed as it's just Mellin convergence at 2.


AlexKontorovich (approved):
Thanks!!

PR #245, opened 2025-03-04
add \end{lemma}

teorth:
Fixes a formatting issue with the previous PR in which lemmas were not being closed properly


AlexKontorovich (approved):
Thanks

PR #246, opened 2025-03-05
golf SmoothedChebyshevDirichlet_aux_tsum_integral

FLDutchmann:
The big lesson here was to avoid `ENNReal.ofReal` until the last possible moment. It's only conditionally multiplicative, so doesn't behave nicely with `simp_rw`. 

The abstract theory of `NNReal` and `ENNReal` is actually quite nice - it's helpful to keep using `enorm` and `nnnorm` until you have to break free of them to use some statement about the complex numbers. 

cc @Aaron1011


AlexKontorovich (approved):
Thanks!

PR #247, opened 2025-03-08
Ak work

AlexKontorovich:


PrimeNumberTheoremAnd/MediumPNT.lean
```diff
@@ -120,45 +142,49 @@ lemma SmoothedChebyshevDirichlet_aux_tsum_integral {SmoothingF : ℝ → ℝ}
...
+  have abs_two : ∀ a : ℝ, ∀ i : ℕ, ‖(i : ℂ) ^ ((σ : ℂ) + ↑a * I)‖₊ = i ^ σ := by
     intro a i
     simp_rw [← norm_toNNReal]
-    norm_cast
-    rw [norm_natCast_cpow_of_re_ne_zero _ (by simp)]
+    -- norm_cast
+    rw [norm_natCast_cpow_of_re_ne_zero _ (by simp only [add_re, ofReal_re, mul_re, I_re, mul_zero,
+      ofReal_im, I_im, mul_one, sub_self, add_zero, ne_eq]; linarith)]
     simp only [add_re, re_ofNat, mul_re, ofReal_re, I_re, mul_zero, ofReal_im, I_im, mul_one,
       sub_self, add_zero, rpow_two, Real.toNNReal_of_nonneg <| sq_nonneg (i : ℝ), Nat.cast_pow]
```
ajirving: You can finish the proof with
```suggestion
      sub_self, add_zero, Real.toNNReal_of_nonneg <| rpow_nonneg (y:= σ) (x:= i) (by linarith)]
    norm_cast
```
and coersion_hell_aux is unneded.

ajirving: Also, we could avoid these simp calls with lots of trivial facts about complex numbers if we extracted lemmas like 
```
lemma foo (a b : ℝ) : (a + b * I).re = a := by simp 
```
we often need versions of this throughout the project.

AlexKontorovich: Thanks!! Fixed

PR #261, opened 2025-06-23
logDeriveResidue type change

maksym-radziwill:
Reverted the conclusion of logDerivResidue to be in terms of filters.


AlexKontorovich (commented):
Thanks!

PR #263, opened 2025-06-24
Added open set assumption to logDerivResidue

maksym-radziwill:
Added an open set assumption to logDerivResidue


PrimeNumberTheoremAnd/ZetaBounds.lean
```diff
@@ -57,13 +58,33 @@ theorem riemannZetaResidue :
...
   If $f$ is holomorphic in a neighborhood of $p$, and there is a simple pole at $p$, then $f'/f$ has a simple pole at $p$ with residue $-1$:
   $$ \frac{f'(s)}{f(s)} = \frac{-1}{s - p} + O(1).$$
 \end{theorem}
 %%-/
-theorem logDerivResidue {f : ℂ → ℂ} {p : ℂ} {U : Set ℂ} (holc : HolomorphicOn f (U \ {p}))
+
+/- The set should be open so that f'(p) = O(1) for all p ∈ U -/
+
+theorem logDerivResidue {f : ℂ → ℂ} {p : ℂ} {U : Set ℂ}
+    (U_is_open : IsOpen U)
```
AlexKontorovich, line 86: You should be able to get something like this from `U_in_nhds` -- this will give you that there is an open set contained in `U` which contains `p`...

AlexKontorovich, line 86:
```suggestion
```

AlexKontorovich (approved):
I'll just push this and we can fix whatever we need to later. Thanks!!

PR #267, opened 2025-06-24
Task 4 is finished

ElifUskuplu:
@shiyue-li and I finished the task 4, namely, proving `riemannZetaLogDerivResidue`. We also suggest to add cify tactic for further use.


AlexKontorovich:
It looks like it didn't pass CI?

AlexKontorovich (approved):
Thanks!!

PR #275, opened 2025-06-25
V8

maksym-radziwill:
Added logDerivResidue with non-vanishing assumption, 
riemannZetaLogDerivResidue now broken due to new assumption (has a sorry in it).


AlexKontorovich (approved):
Thanks!

PR #287, opened 2025-06-25
2/3 sorrys on SmoothedChebyshevPull1 complete

prestontranbarger:


PrimeNumberTheoremAnd/Wiener.lean
```diff
@@ -370,7 +370,7 @@ for all $u \in \R$.
 %%-/
 
 /-%%
-\begin{proof}\uses{prelim-decay, prelim-decay-3} Should follow from previous lemmas.
+\begin{proof}\uses{prelim-decay, prelim-decay-3, decay} Should follow from previous lemmas.
```
AlexKontorovich:
```suggestion
\begin{proof}\uses{prelim-decay, prelim-decay-3} Should follow from previous lemmas.
```

PrimeNumberTheoremAnd/Wiener.lean
```diff
@@ -454,7 +454,7 @@ lemma W21.integrable_fourier (ψ : W21) (hc : c ≠ 0) :
   apply Eventually.of_forall (fun x => h _)
 
 /-%%
-\begin{proof}\leanok\uses{decay-alt} From two integration by parts we obtain the identity
+\begin{proof}\leanok From two integration by parts we obtain the identity
```
AlexKontorovich:
```suggestion
\begin{proof}\leanok\uses{decay-alt} From two integration by parts we obtain the identity
```

PR #288, opened 2025-06-25
riemannZetaLogDerivResidue fix and auxiliary lemmas

ElifUskuplu:
I fixed the `sorry`s in `riemannZetaLogDerivResidue` and added two another results. The former is a useful fact I think we can use more later, the latter is another version of `logDerivResidue`.


AlexKontorovich (approved):
Thank you!!

PR #312, opened 2025-06-28
proof of ResidueOfTendsTo

ElifUskuplu:
I followed the same proof in the blueprint, but let me know if I can make it shorter.


AlexKontorovich (approved):
Thank you!!

PR #319, opened 2025-07-01
Correct reference to deprecated name

emo916math:
Corrected the reference to `gt_of_ge_of_gt` to the updated name `lt_of_le_of_lt'`.


AlexKontorovich (approved):
Thanks!!

PR #322, opened 2025-07-02
verticalIntegral_split_three_finite

ajirving:
Gives a short proof of verticalIntegral_split_three_finite.  Also removes an unused havve from SmoothedChebyshevPull1.


AlexKontorovich (approved):
Thanks!!

PR #323, opened 2025-07-02
I2 bound

emo916math:
Finished I2Bound, up to a good bound for ζ'/ζ on horizontal segments crossing σ = 1, away from the pole.

We need such a bound. We have `LogDerivZetaBndUniform` only for σ < 1; we have `dlog_riemannZeta_bdd_on_vertical_lines'`, where the constant is not explicit; and we have `dlog_riemannZeta_bdd_on_vertical_lines_generalized`, where the constant is ζ'/ζ(σ), which is not good when σ ≈ 1.


AlexKontorovich (approved):
Thanks!!

PR #325, opened 2025-07-03
Pull2

ajirving:
Fills the remaining sorry in pull2 and fixes a couple of depprications which seem to have been reintroduced by the last PR.

Note that pull2 is not actually complete because pull2_aux1 is still unproved.  That won't be completely trivial because we're asking for continuity at a point on the boundary of the rectangle of holomorphy.


AlexKontorovich (approved):
Thanks!

PR #326, opened 2025-07-04
Symmetry: Sketched the proof that ζ(s) = conj (ζ (conj s).

emo916math:
ζ(s) = conj (ζ (conj s).

This is a useful theorem in its own right, and it also streamlines the passage from I_1 to I_9, etc. in the PNT bounding.

It remains to show that if f is an analytic function, so is conj (f (conj s). In fact the derivative is conj (f' (conj s), which we will need in dealing with ζ'/ζ.


AlexKontorovich (approved):
Thanks!

PR #332, opened 2025-07-05
Proved I8Bound using symmetry.

emo916math:
This was a test case; the other bounds I6..I9 can likely be proved similarly.


AlexKontorovich (approved):
Thanks!!

PR #336, opened 2025-07-10
Finished I3 and I7 Bounds

stevenecreech:
I finished the formalization for the I3 and I7 integrals. I shall remark that I wrote a helper lemma that I called lemma log_pow_over_xsq_integral_bounded which was an integral that I used in both to help with the symmetry. Furthermore, I will remark that I think I had to import pi_bounds from Mathlib, and I also had to increase the max-heartbeat counter in order for the formalization to go through, but they are done :)


ajirving:
The reason the blueprint is failing to build is that you're missing a \end{proof} on the proof of i8bound which is just before your new lemma.

AlexKontorovich (approved):
Thanks!!

PR #340, opened 2025-07-12
Work on holomorphy results

ajirving:
Various small improvements to the LogDerivZetaHolc results so that MediumPNT can now easily call pull1:
* Moved a couple of results about zeta from MediumPNT.lean to ZetaBounds.lean (and shortened the proofs).
* a general lemma that to get holomorphy of zeta'/zeta on a set we just need to avoid 1 and be zero free.  This significantly shortens the following results.
* Make the bound in NoZerosInBox be inclusive.  This further simplifies the holomorphy results because we can just use T = 3 rather than needing the result at 4 too.
* Rework LogDerivZetaHolcLargeT to hold for all T >= 3 rather than just sufficiently large (by reducing the constant A). Also changed this to use closed intervals to match the MediumPNT requirement.
* The new HolcLargeT can now be used in MediumPNT to implement pull1.

### DIFF
--- a/PrimeNumberTheoremAnd/MediumPNT.lean
+++ b/PrimeNumberTheoremAnd/MediumPNT.lean
@@ -6410,17 +6410,22 @@ theorem MediumPNT : ∃ c > 0,
 
   let T : ℝ := sorry
   have T_gt_3 : 3 < T := sorry
-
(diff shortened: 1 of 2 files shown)


AlexKontorovich (approved):
Thanks!

PR #342, opened 2025-07-14
Uniform version of all IBounds except I4, I5, I6

maksym-radziwill:
Here are the changes that make the bounds uniform.

The statements for I4, I5, I6 need to be changed even more since we need to be able to pick the same \sigma_2 in all of them. For I5 there is an extra issue that we might want to stay away from 1, hence the extra \eta > 0. I actually don't like the proposed new statement for I5, hopefully you have a better one...


ajirving:
I wonder if a cleaner alternative would be to pass the values of A and $\sigma_2$ into these theorems along with the required holomorphy and bound hypotheses on $\zeta$.  The results on zeta can then just be used in the MediumPNT proof to obtain the constants and hypotheses and pass them to all the integral bounds.

ajirving:
I've been experimenting and I don't think its any more difficult to make the bounds take the necessary bounds as arguments.  I'll make a PR with what i have in case its helpful.

PR #344, opened 2025-07-20
Passing zeta bounds

ajirving:
Here's my version of passing in the properties of zeta we need to the various I*bounds.  Please feel free to either merge in all of it or just extract parts you find helpful.  Specifically we have:
* Definitions LogDerivZetaHasBound and LogDerivZetaHoloSmall which capture the properties we want.
* Versions of I1235789bound which take these as arguments. I didn't do I46 because I see they're being worked on in #343.
* Updates MediumPNT to use all of these bounds.  I also did some reordering in MediumPNT to put the constants we need at the start and to filter_upwards in X.


AlexKontorovich (approved):
Thanks!

PR #350, opened 2025-07-22
Main term

ajirving:
Adds an explicit version of MellinOfSmooth1c and uses it to handle the main term in MediumPNT>


AlexKontorovich (approved):
Thanks!

PR #359, opened 2025-07-30
Mellin from Mathlib

ajirving:
Use the Mellin transform, its inverse and Mellin inversion directly from Mathlib rather than defining our own versions.  There are a number of mostly minor changes to support the slightly different definition of Mellin. The PerronFormula.lean file is no longer imported but I haven't deleted it.


PrimeNumberTheoremAnd/MellinCalculus.lean
```diff
@@ -1,6 +1,6 @@
 import PrimeNumberTheoremAnd.Auxiliary
 import Mathlib.Analysis.MellinInversion
-import PrimeNumberTheoremAnd.PerronFormula
+import PrimeNumberTheoremAnd.Wiener
```
AlexKontorovich: Why does this file need to import Wiener?..

AlexKontorovich (approved):
Thanks!!

PrimeNumberTheoremAnd/MellinCalculus.lean
ajirving: The proof of SmoothExistence uses smooth_urysohn_support_Ioo from Wiener.  Is there somewhere better we could move that to?

AlexKontorovich: Yes I think we should move that to its own file, and import that in both Wiener and MellinCalculus. (Come to think of it, the actual existence shouldn't be needed until MediumPNT, no? We can move SmoothExistence to the new file...)

ajirving: Its now in SmoothExistence.lean and imports have been fixed.

PR #373, opened 2025-08-15
chore: split a lemma out of MediumPNT

Ruben-VandeVelde:
The proof is very long and slow; this makes a small improvement.


AlexKontorovich (approved):
Thanks!

PR #376, opened 2025-08-23
Zeta tidy

ajirving:
Various tidying of zeta bounds as discussed on Zulip:
* Updates the blueprint to more accurately track the Lean proof.
* Removes LogDerivZetaBndUniform as it was unused and weaker than the UNif version.
* Keep only the stronger versions of LogDerivZetaBnd
* Moves the unused LogDerivZetaBndAlt into an unused file.  Maybe it isn't worth keeping at all in which case I can remove that file.
* Moves LogDerivZetaBndUnif and its dependencies into ZetaBounds.lean and simplifies some of the proofs.  In particular significant shortening of triv_bound_zeta.


PrimeNumberTheoremAnd.lean
```diff
@@ -11,3 +11,4 @@ import «PrimeNumberTheoremAnd».MediumPNT
 import «PrimeNumberTheoremAnd».StrongPNT
 import «PrimeNumberTheoremAnd».Wiener
 import «PrimeNumberTheoremAnd».ZetaBounds
+import «PrimeNumberTheoremAnd».ZetaBoundsUnused
```
AlexKontorovich, line 14: Why is this getting imported if it's not getting used? Thanks!

ajirving: I thought it was a good idea to make sure it still compiles when we make further changes to the code. Otherwise if it breaks we probably wouldn't notice. I'm happy to remove the import if you prefer, there are other files not being imported like PerronFormula.

AlexKontorovich (approved):
Thanks!

PR #378, opened 2025-08-30
Zeta golf

ajirving:
Some more tidying in ZetaBounds.lean.  NO significant mathematical changes but various places where we can use existing Mathlib results and some shortening of proofs.


AlexKontorovich (approved):
Thanks!

PR #385, opened 2025-09-07
Pull1residue

ajirving:
shortening the proof of SmoothedChebyshevPull1.   In particular I extracted 2 smaller theorems which give a more useful version of the residue theorem (ResidueTheoremOnRectangleWithSimplePole') where the residue condition is local to the pole.  Perhaps these new results should be moved to ResidueCalcOnRectangles.lean along with the results from ZetaBounds on which they depend.


ajirving:
This will probably have some conflicts with #384, I can fix those if you merge that first.

AlexKontorovich:
> shortening the proof of SmoothedChebyshevPull1. In particular I extracted 2 smaller theorems which give a more useful version of the residue theorem (ResidueTheoremOnRectangleWithSimplePole') where the residue condition is local to the pole. Perhaps these new results should be moved to ResidueCalcOnRectangles.lean along with the results from ZetaBounds on which they depend.

Great, thanks! Yes these are general facts about holomorphic functions, so don't belong in the MediumPNT file - please move them. Thank you!

AlexKontorovich (approved):
Thanks!

PR #391, opened 2025-10-05
Section 3.2 upstream

prestontranbarger:
I've completed section 3.2 which gives the upper bound
|zeta(s+3/2+it)| leq 7+2|t|
for |s| leq 1 in C and |t| geq 3 in R.


AlexKontorovich (approved):
Thanks!

PR #402, opened 2025-10-14
Shortening smoothedChebyshevClose_aux

ajirving:
Some simplifications, trying to replace the long series of haves by something more structured using calc.   This proof is still very long but this is some improvement.


AlexKontorovich (approved):
Thanks!

PR #416, opened 2025-11-24
Main

teorth:


teorth (commented):
Finish the proof of pn_asymptotic, thus filling in some sorries in Consequences.lean

PR #433, opened 2025-12-19
Update proof of vonMangoldt_cheby

ajirving:
Use the Chebyshev estimate in Mathlib rather than Brun-Tichmarsh.


AlexKontorovich (approved):
Thanks!

PR #468, opened 2026-01-07
feat(Erdos392): fix `Params` and prove `initial.waste`

pitmonticone:
The negation of the misformalised version was autoformalised by @Aristotle-Harmonic, then I manually proved the fixed version of `Params.initial.waste` with the missing hypothesis `hM: M > 1` added to the `Params` structure.


teorth:
Looks good.  In general I think the strategy of creating structures such as `Params` to hold all the side conditions that will inevitably come up will help keep the global proof structure manageable.  (I expect this to become more of an issue when we dive deeper into the explicit analytic number theory estimates, where there are actually a huge number of side conditions and relations between constants and parameters that are often left implicit, or verified only for a specific hard-coded set of parameters.)

PR #484, opened 2026-01-08
prod_q_ge done

AlexKontorovich:


PrimeNumberTheoremAnd/Lcm.lean
```diff
@@ -742,7 +742,46 @@ blueprint_comment /--
...
+    conv =>
+      enter [1, 2, 2, 1, 2]
+      equals 3 => norm_cast
+    conv =>
+      enter [2, 2, 2, 1, 2]
+      equals 1 => norm_cast
+    conv =>
+      enter [2, 1, 2, 2, 1, 2]
+      equals 2 => norm_cast
+    ring]
```
AlexKontorovich, line 765: This surely can be done better with a call to `Prod.equiv` of some kind. This was faster to do, with `conv?` calls and point and clicking; but of course looks very ugly...

PR #487, opened 2026-01-08
feat(Erdos392): prove `lower_score_3`

pitmonticone:
This PR adds proofs autoformalised by @Aristotle-Harmonic with substantial manual refactoring.


teorth:
Thanks!  I guess there's no obvious way to make the proof much shorter than this.  I'm curious as to the workflow here - did you need to manually identify many of the key sublemmas which you then delegated to Aristotle to prove?

Issue #512, opened 2026-01-09
[LCM]: Verification of criterion for large n


teorth:
## Description

Verifies the criterion for $$n$$ large enough.  Everything except `h_crit` is already formalized; this should follow from combining the previous few lemmas.

## Blueprint

[Link](https://alexkontorovich.github.io/PrimeNumberTheoremAnd/blueprint/tertiary-chapter.html#prop:ineq-holds-large-n)

## Zulip

[Link](https://leanprover.zulipchat.com/#narrow/channel/423402-PrimeNumberTheorem.2B/topic/Explicit.20estimates.2C.20outstanding.20tasks.20v2.2E)

PR #539, opened 2026-01-11
Correct the statement of `prod_epsilon_ge`.

CoolRmal:
We should divide `1 + 1/89693` instead of multiplying it.

I used `* (1 / (1 + 1/89693))` because this is the formulation used in `inv_n_add_sqrt_ge`, which makes it easy for me to finish #512


CoolRmal:
3.36683 works for both `prod_epsilon_ge` and `final_comparison`. I'll update my PR now.

PR #522, opened 2026-01-09
Ak work

AlexKontorovich:
solved with help from AlphaProof

Closes #511

PR #579, opened 2026-01-12
cleanup numerical calculation proof

b-mehta:
I had claimed and finished this `sorry` on Zulip, but it was implemented anyway in #522. Since my proof is quite a bit cleaner, I add it anyway.


pitmonticone (approved):
Thanks!

PR #580, opened 2026-01-12
cleanup verification of large n criterion

b-mehta:
I had claimed this goal on zulip, but it was proved in #544 by someone else. Since my proof is shorter and faster, I'm PRing it anyway.


pitmonticone (approved):
Thanks!

PR #587, opened 2026-01-13
lemma_abadeuleulmit1 fix

AlexKontorovich:


PrimeNumberTheoremAnd/ZetaAppendix.lean
```diff
@@ -877,8 +877,8 @@ after reindexing the second sum. Regrouping terms again, we obtain our equation.
   (latexEnv := "lemma")
   (discussion := 569)]
 theorem lemma_abadeuleulmit1 {z : ℂ} (hz : ¬∃ n : ℤ, z = n) :
-    (π / Complex.sin (π * z) : ℂ) =
-      (1 / z : ℂ) + ∑' n : ℤ, (-1) ^ n * ((1 / (z - n) : ℂ) + (1 / (z + n) : ℂ)) := by
+    (π / Complex.sin (π * z)) =
+      (1 / z) + ∑' (n : {m : ℤ // m > 0}), (-1) ^ (n : ℤ) * ((1 / (z - n) : ℂ) + (1 / (z + n) : ℂ)) := by
```
CoolRmal, line 881: Actually this shouldn't be indexing over `ℤ` right?
I think we want either

```
(1 / z) + ∑' (n : ℕ+), (-1) ^ (n : ℤ) * ((1 / (z - n) : ℂ) + (1 / (z + n) : ℂ))
```
or 

```
(1 / z) + ∑' (n : ℕ), (-1) ^ (n : ℤ) * ((1 / (z - (n + 1)) : ℂ) + (1 / (z + (n + 1)) : ℂ))
```

The first one correspond to [cot_series_rep](https://leanprover-community.github.io/mathlib4_docs/Mathlib/Analysis/SpecialFunctions/Trigonometric/Cotangent.html#cot_series_rep), and the second one corresponds to [cot_series_rep'](https://leanprover-community.github.io/mathlib4_docs/Mathlib/Analysis/SpecialFunctions/Trigonometric/Cotangent.html#cot_series_rep').

teorth: I think mathematically there is little difference between summing over positive integers or over positive natural numbers; it's a question of which spelling will be more convenient in Lean.  It's only two or three lines of some boring casting tactics or something to change one to the other (and all the autoformalizers will be able to handle it), so I guess it is up to the human doing the formalizing to exercise their aesthetic judgement as to what the best spelling is.

PR #622, opened 2026-01-15
chore(FKS2): correct statement of `lemma_10c`

arajasek:
The original FKS2 paper appears to have a typo in the statement of this lemma -- it is trivial to verify that the correct condition is `\sqrt{\log x} < -2b/c`.

### DIFF
--- a/PrimeNumberTheoremAnd/FKS2.lean
+++ b/PrimeNumberTheoremAnd/FKS2.lean
@@ -106,12 +106,13 @@ theorem lemma_10b {a b c : ℝ} (ha : a > 0) (hc : c > 0) (hb : b ≥ -c ^ 2 / (
   "fks2-lemma-10c"
   (title := "FKS2 Lemma 10c")
   (statement := /--
-  If $c>0$, $g(0,b,c,x)$ decreases with $x$ for $\sqrt{\log x} > -2b/c$. -/)
+  If $c>0$, $g(0,b,c,x)$ decreases with $x$ for $\sqrt{\log x} < -2b/c$. -/)
   (proof := /-- We apply Lemma \ref{fks2-lemma-10-substep-2}. If $a = 0$, it is negative when $u < \frac{-2b}{c}$.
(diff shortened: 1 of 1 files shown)


PrimeNumberTheoremAnd/FKS2.lean
```diff
@@ -97,12 +97,13 @@ theorem lemma_10b {a b c : ℝ} (ha : a > 0) (hc : c > 0) (hb : b ≥ -c ^ 2 / (
...
   (statement := /--
-  If $c>0$, $g(0,b,c,x)$ decreases with $x$ for $\sqrt{\log x} > -2b/c$. -/)
+  If $c>0$, $g(0,b,c,x)$ decreases with $x$ for $\sqrt{\log x} < -2b/c$. -/)
   (proof := /-- We apply Lemma \ref{fks2-lemma-10-substep-2}. If $a = 0$, it is negative when $u < \frac{-2b}{c}$.
+  Note: this lemma is mistyped as $\sqrt{\log x} > -2b/c$ in \cite{FKS2}.
  -/)
   (latexEnv := "lemma")]
 theorem lemma_10c {b c : ℝ} (hc : c > 0) :
-    StrictAntiOn (g_bound 0 b c) (Set.Ioi (exp ((-2 * b / c) ^ 2))) := sorry
+    StrictAntiOn (g_bound 0 b c) (Set.Iio (exp ((-2 * b / c) ^ 2))) := sorry
```
arajasek: I wasn't sure whether this, or `Set.Ioo 1 (exp ((-2 * b / c) ^ 2))` was more appropriate.

teorth: Well, sqrt(log x) is likely to behave very weirdly for x below 1, right? So most likely the claim will not be true for x < 1, and we only plan to use this above 1, so the Set.Ioo form may indeed be better.

arajasek: Fixed up, thank you

PR #687, opened 2026-01-18
fix(Rosser): add parentheses around integrals

CoolRmal:
Otherwise lean will interpret `∫ y in 2..x, f y / log y + 2 * f 2 / Real.log 2` as

$$\int_2^x(\frac{f(y)}{\log y}+\frac{2f(2)}{\log 2})dy$$ instead of

$$\int_2^x\frac{f(y)}{\log y}dy + \frac{2f(2)}{\log 2}$$.


CoolRmal:
@teorth I just noticed that the same error appears in other files as well. I’m happy to open another PR to address these, but it might be easier for you to handle them directly. Let me know what you prefer.

PR #693, opened 2026-01-18
feat(BKLNW): prove `prop_3_sub_3`

### LINKED ISSUE #633: [BKLNW]: f(2^n) = 1 + u_n (Prop. 3, substep 3)
## Description

Routine substitution

## Blueprint

[Sublemma 9.3.4](https://alexkontorovich.github.io/PrimeNumberTheoremAnd/blueprint/secondary-chapter.html#bklnw-prop-3-sub-3)

## Zulip

[Link](https://leanprover.zulipchat.com/#narrow/channel/423402-PrimeNumberTheorem.2B/topic/Starting.20on.20BKLNW)

jdominpa:
I've changed the floor function in the definition of `f` to `Nat.floor` instead of `Int.floor`. There may be a shorter way of doing the proof since I'm new to Lean.

Closes #633

### DIFF
--- a/PrimeNumberTheoremAnd/BKLNW.lean
+++ b/PrimeNumberTheoremAnd/BKLNW.lean
@@ -167,8 +167,8 @@ noncomputable def u (n : ℕ) : ℝ := ∑ k ∈ Finset.Icc 4 n, 2^((n/k:ℝ) - 
   (discussion := 633)]
 theorem prop_3_sub_3 (n : ℕ) (hn : n ≥ 3) : f (2^n) = 1 + u n := by
   have sum_bound : ⌊ (log (2 ^ n)) / (log 2) ⌋₊ = n := by norm_num
-  rw [f, u, sum_bound, ← Finset.add_sum_Ioc_eq_sum_Icc hn, ← Finset.Ioc_eq_Icc, Nat.cast_ofNat,
-    sub_self, rpow_zero]
+  rw [f, u, sum_bound, ← Finset.add_sum_Ioc_eq_sum_Icc hn,
+    ← Finset.Icc_add_one_left_eq_Ioc, Nat.cast_ofNat, sub_self, rpow_zero]
   congr with k
   rw [← rpow_natCast _ n, ← rpow_mul (by norm_num)]
   field_simp


PrimeNumberTheoremAnd/BKLNW.lean
```diff
@@ -157,7 +157,16 @@ noncomputable def u (n : ℕ) : ℝ := ∑ k ∈ Finset.Icc 4 n, 2^((n/k:ℝ) -
   (proof := /-- Clear. -/)
   (latexEnv := "sublemma")
   (discussion := 633)]
-theorem prop_3_sub_3 (n : ℕ) : f (2^n) = 1 + u n := by sorry
+theorem prop_3_sub_3 (n : ℕ) (hn : n ≥ 3) : f (2^n) = 1 + u n := by
+  have sum_bound : ⌊ (log (2 ^ n)) / (log 2) ⌋₊ = n := by norm_num
+  rw [f, u, sum_bound]
+  rw [← Finset.add_sum_Ioc_eq_sum_Icc (by linarith), ← Finset.Ioc_eq_Icc]
```
teorth: one could combine these two `rw`s into one, I think (this doesn't actually speed things up much, but is generally accepted practice when writing Lean).

jdominpa: I see, I'll change them to a single `rw`.

PrimeNumberTheoremAnd/BKLNW.lean
```diff
@@ -157,7 +157,15 @@ noncomputable def u (n : ℕ) : ℝ := ∑ k ∈ Finset.Icc 4 n, 2^((n/k:ℝ) -
   (proof := /-- Clear. -/)
   (latexEnv := "sublemma")
   (discussion := 633)]
-theorem prop_3_sub_3 (n : ℕ) : f (2^n) = 1 + u n := by sorry
+theorem prop_3_sub_3 (n : ℕ) (hn : n ≥ 3) : f (2^n) = 1 + u n := by
+  have sum_bound : ⌊ (log (2 ^ n)) / (log 2) ⌋₊ = n := by norm_num
+  rw [f, u, sum_bound, ← Finset.add_sum_Ioc_eq_sum_Icc (by linarith), ← Finset.Ioc_eq_Icc]
```
CoolRmal: I don't know why `Finset.Ioc_eq_Icc` is introduced in ZetaBounds.lean. Maybe one should delete that lemma in ZetaBounds.lean and use [Finset.Icc_add_one_left_eq_Ioc_of_not_isMax](https://leanprover-community.github.io/mathlib4_docs/Mathlib/Algebra/Order/Interval/Finset/SuccPred.html#Finset.Icc_add_one_left_eq_Ioc_of_not_isMax)

PrimeNumberTheoremAnd/BKLNW.lean
```diff
@@ -157,7 +157,15 @@ noncomputable def u (n : ℕ) : ℝ := ∑ k ∈ Finset.Icc 4 n, 2^((n/k:ℝ) -
...
+  congr
+  · norm_num
+  ext k
+  calc (2 ^ n : ℝ) ^ (1 / ↑k - 1 / 3 : ℝ)
+    _ = 2 ^ (n * (1 / ↑k - 1 / 3 : ℝ)) := by rw [← rpow_natCast _ n, ← rpow_mul (by norm_num)]
+    _ = _ := by field_simp
```
CoolRmal:
```suggestion
  rw [f, u, sum_bound, ← Finset.add_sum_Ioc_eq_sum_Icc hn, ← Finset.Ioc_eq_Icc, Nat.cast_ofNat,
    sub_self, rpow_zero]
  congr with k
  rw [← rpow_natCast _ n, ← rpow_mul (by norm_num)]
  field_simp
```

PrimeNumberTheoremAnd/BKLNW.lean
jdominpa: Should I replace `Finset.Ioc_eq_Icc` in case it gets deleted from ZetaBounds.lean in the future? Also, what's the difference between `Finset.Icc_add_one_left_eq_Ioc_of_not_isMax` and `Finset.Icc_add_one_left_eq_Ioc`?

CoolRmal: I think we shouldn't duplicate lemmas if we already have them in Mathlib, so my suggestion for you would be
1. Delete `Finset.Ioc_eq_Icc` in ZetaBounds.lean.
2. Fix any issues caused by deleting this lemma in this repository (probably by replacing `Finset.Ioc_eq_Icc` with `Finset.Icc_add_one_left_eq_Ioc_of_not_isMax` or `Finset.Icc_add_one_left_eq_Ioc`).

Also, I am not an official reviewer, so please take these comments with a grain of salt.

Regarding your last question, `Finset.Icc_add_one_left_eq_Ioc_of_not_isMax` is used when an element is not the maximum, and `Finset.Icc_add_one_left_eq_Ioc` is used when the underlying type has no maximum. In this case, it is probably easier to use `Finset.Icc_add_one_left_eq_Ioc` as I believe Lean is smart enough to infer that `ℕ` has [NoMaxOrder](https://leanprover-community.github.io/mathlib4_docs/Mathlib/Order/Max.html#NoMaxOrder).

teorth:
According to the CI, there is still an unresolved goal:
```
  error: PrimeNumberTheoremAnd/BKLNW.lean:200:71: unsolved goals
  n : ℕ
  hn : n ≥ 9
  ⊢ f (2 ^ (n + 1)) < f (2 ^ n)
```

pitmonticone:
Done

pitmonticone (approved):
LGTM.

PR #697, opened 2026-01-19
feat(FKS2): prove `lemma_10_substep`

Steven-Rossi:


Steven-Rossi:
Not sure if we should merge yet, changing g_bound to use \R makes lemma_10c break, im investigating it right now

pitmonticone:
@Steven-Rossi It's probably because it was missing `(hc : c > 0)` and the change from `(log x) ^ (1/2)` to `sqrt (log x)` makes it quite a bit harder to work with.

pitmonticone (approved):
Now it looks good to me.

Issue #759, opened 2026-01-23
[LOG]: Show that li(2) ~ 1.0451 (The Ramanujan-Soldner constant)


teorth:
## Description

This may require first rewriting this principal value integral as a more convergent integral, and then performing some numerical integration.  Will likely overlap with #758.

## Blueprint

[Theorem 9.1.1](https://alexkontorovich.github.io/PrimeNumberTheoremAnd/blueprint/secondary-chapter.html#Ramanujan-Soldner-constant)

## Zulip

[Link](https://leanprover.zulipchat.com/#narrow/channel/423402-PrimeNumberTheorem.2B/topic/Miniproject.3A.20prove.20that.20li.28x.29.3DLi.28x.29.2Bli.282.29.20and.20li.282.29~.201.2E0451)

PR #774, opened 2026-01-25
feat(Li2Bounds): prove li(2) (Ramanujan-Soldner constant) bounds via LeanCert verified integration

alerad:
Uses LeanCert's Li2Verified module for machine-checked bounds: 1.039 ≤ li(2) ≤ 1.06

What's proven (no sorries):

li2_symmetric_eq_li2: symmetric form = principal value
li.two_approx_weak: 1.039 ≤ li(2) ≤ 1.06
All bounds in Li2Bounds.lean
Architecture: Li2Bounds.lean is self-contained (stable, won't need rebuilds)

Addresses #759


lake-manifest.json
```diff
@@ -95,7 +105,7 @@
    "type": "git",
    "subDir": null,
    "scope": "leanprover-community",
-   "rev": "b25b36a7caf8e237e7d1e6121543078a06777c8a",
+   "rev": "45926c76c3e0d876e39b3e9e5e77080fef0b10ba",
```
ajirving: Why has the revision of batteries changed?  Shouldn't we be staying on the tagged v4.27.0 revision?

teorth:
It might be worth taking some notes on these experiences and then summarizing them somewhere on the Zulip.  I must imagine that these sort of back end integration issues are ones that other developers of Lean imports are going to face, and having some guideposts on potential things to watch out for could be very helpful.

alerad:
Thanks for raising this, it's actually something that's been on my mind throughout this work as well.

I do have some ideas for reducing the partition count by improving how the Taylor-model integration is handled, rather than relying on uniform subdivision. I haven't implemented these yet because I wanted to first get a fully correct, blueprint-aligned proof integrated before tuning performance.

This build time isn't inherent to the approach, it's an artifact of conservative choices made for this initial version. It's my top priority once the PR settles. I expect to be able to speed it up by an order of magnitude.

teorth:
I think this is a good compromise: retains the numerical calculations, but doesn't hook them up completely to the main repository in order to keep the compilation at reasonable sizes.  At the very end of the project one could perform the final hookups to get something closer to an end-to-end theorem, or one couuld leave them disconnected and rely on some metaframework above the Lean code to give a (slightly less rigorous) certificate than full Lean verification.

teorth:
Finally got it to build in a reasonable time!  I think we can experiment with adopting it now.

Issue #873, opened 2026-02-04
[IK]: prove `sum_moebius_pmul_eq_prod_one_sub`


AlexKontorovich:
## Description

Prove that `∑ d | n, μ(d) * g(d) = ∏ p | n, (1 - g(p))` holds for multiplicative functions

## Blueprint

[](https://alexkontorovich.github.io/PrimeNumberTheoremAnd/blueprint/IK-chapter.html#sum_moebius_pmul_eq_prod_one_sub)

## Zulip

TBA

PR #877, opened 2026-02-04
chore(workflows): temporarily downgrade docgen-action

arajasek:
Temporary workaround to get CI passing again.

Also removes the dummy `references.bib` file


arajasek:
Also fixes up some typesetting issues in #873 and others that caused CI to fail.

Issue #844, opened 2026-02-01
[CHEBYSHEV]: Clean upper bound for psi


teorth:
## Description

The previous bound handles large x; some numerical computation needed to handle smaller x

## Blueprint

[Theorem 9.2.3](https://alexkontorovich.github.io/PrimeNumberTheoremAnd/blueprint/secondary-chapter.html#psi-upper-clean)

## Zulip

TBD

PR #1007, opened 2026-02-17
feat(Chebyshev): prove psi_upper_clean (#844)

alerad:
## Summary
- Proves `psi_num_2` (`ψ(x) ≤ 1.11x` for `0 < x ≤ 11723`) by importing LeanCert’s certified bound [`PNT_PsiBounds.psi_le_mul_11723`](https://github.com/alerad/leancert/blob/596bc8fc6ed55f13287f2052db84274c300f6a7b/LeanCert/Examples/PNT_PsiBounds.lean).
- Proves `psi_upper_clean` (`ψ(x) ≤ 1.11x` for `x > 0`) via strong induction, using `psi_diff_upper` plus a logarithmic bound with `interval_decide`.
- Pins `leancert` to `v4.28.0-rc1` (manifest rev `596bc8fc6ed55f13287f2052db84274c300f6a7b`) to include [`ChebyshevPsi`](https://github.com/alerad/leancert/blob/596bc8fc6ed55f13287f2052db84274c300f6a7b/LeanCert/Engine/ChebyshevPsi.lean) / [`PNT_PsiBounds`](https://github.com/alerad/leancert/blob/596bc8fc6ed55f13287f2052db84274c300f6a7b/LeanCert/Examples/PNT_PsiBounds.lean).

### DIFF
--- a/PrimeNumberTheoremAnd/Chebyshev.lean
+++ b/PrimeNumberTheoremAnd/Chebyshev.lean
@@ -6,7 +6,7 @@ import Mathlib.NumberTheory.Chebyshev
 import Mathlib.Tactic.NormNum.BigOperators
 import PrimeNumberTheoremAnd.LogTables
 import PrimeNumberTheoremAnd.SecondaryDefinitions
-import LeanCert.Examples.PNT_PsiBounds
+import LeanCert.Engine.ChebyshevPsi
 import LeanCert.Tactic.IntervalAuto
 
 blueprint_comment /--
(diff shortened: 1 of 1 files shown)


alerad:
Thanks, details are also posted on the Chebyshev Zulip topic

For `psi_num_2`, yes: LeanCert verifies `ψ(N) ≤ 1.11N for all N = 1..11723` directly via `checkAllPsiLeMulWith` (an O(N) incremental checker) and discharges it with one native_decide; no sparse checkpoint ladder is used.

This is not the earlier CI-only/sorry-interface pattern from heavier cases: PNT imports a fully proved LeanCert theorem here, so the ~20s native_decide is on PNT+'s build path. 

Since it's already fully proved and on the build path, I could also move the verification inline into PNT+ for blueprint visibility. The tradeoff is ~20s on every Chebyshev.lean rebuild vs only on LeanCert bumps. Happy to do whichever you prefer.

LeanCert changelog since last pin posted on the LeanCert Imported topic.

PR #1096, opened 2026-03-04
[StrongPNT]: SetOfZeros, ZeroFactorization, Cf, BlaschkeB, BlaschkeOfZero, DiskBound, and ZerosBound

prestontranbarger:


PrimeNumberTheoremAnd/StrongPNT.lean
```diff
@@ -481,8 +480,35 @@ blueprint_comment /--
     Trivially, $h_\rho(z)$ is analytic at $\rho$ (we have written down the series
     expansion); now note that
     $$h_\rho(\rho)=\sum_{m\leq n}a_n(\rho-\rho)^{n-m}=\sum_{m\leq n}a_n0^{n-m}=a_m\neq 0.$$
-\end{proof}
--/
+  -/)]
+lemma ZeroFactorization {f : ℂ → ℂ} (hfAnalytic : ∀ z ∈ Metric.closedBall (0 : ℂ) 1, AnalyticAt ℂ f z)
```
ajirving: You can write hfAnalytic using AnalyticOnNhd

PrimeNumberTheoremAnd/StrongPNT.lean
```diff
@@ -481,8 +480,35 @@ blueprint_comment /--
...
+    have ρ_mem_ball : ρ ∈ Metric.ball (0 : ℂ) 1 := by
+      rw[mem_ball_iff_norm, sub_zero]
+      linarith[hρ.1]
+    have f_analAt_ρ : AnalyticAt ℂ f ρ := by
+      apply hfAnalytic ρ
+      rw[mem_closedBall_iff_norm]
+      rw[mem_ball_iff_norm] at ρ_mem_ball
+      exact le_of_lt ρ_mem_ball
+    have finiteOrder : analyticOrderAt f ρ ≠ ⊤ := by
+      by_contra orderIsTop
```
ajirving: Maybe shorter to avoid contradiction here by using AnalyticOnNhd.analyticOrderAt_ne_top_of_isPreconnected

PrimeNumberTheoremAnd/StrongPNT.lean
prestontranbarger: done! thanks for the suggestion

PrimeNumberTheoremAnd/StrongPNT.lean
prestontranbarger: done! thanks for the suggestion

PrimeNumberTheoremAnd/StrongPNT.lean
```diff
@@ -500,42 +518,73 @@ blueprint_comment /--
...
--/
+  -/)]
+noncomputable def Cf
+    {r : ℝ} (r_lt_one : r < 1)
+    {f : ℂ → ℂ} (finiteZeros : (SetOfZeros r f).Finite)
+    (hfAnalytic : AnalyticOnNhd ℂ f (Metric.closedBall (0 : ℂ) 1))
+    (hf_neq_zero_at_zero : f 0 ≠ 0)
+    (z : ℂ) : ℂ :=
+  if z_zero : z ∈ SetOfZeros r f then
+    (ZeroFactorization hfAnalytic hf_neq_zero_at_zero r_lt_one z_zero).choose z
```
ajirving: It might be worthwhile giving some thought to how we want to write this definition.  At a minimum I think the first case is written incorrectly, don't you need to divide out by the zeros distinct from z?  Other thoughts, I'm not saying any of them are definitely the right thing to do:
* I don't like having a definition which requires lots of hypotheses.  One option would be to use Mathlib's  meromorphicTrailingCoeffAt which I think gives the same value as ZeroFactorization.
* Rather than assuming the zero set is finite you could use the Mathlib API for divisors of Meromorphic functions and have a finprod over the divisor.

Maybe look at the Mathlib version of Jensen's formula to see some examples of how the above is used.

prestontranbarger: Okay, I agree I wrote the definition incorrectly in the definition. I have resolved it on my end, and I will resolve it in a PR with diskBound as well. I'll take a look at the other two things you pointed out.

Thanks

ajirving: Thinking some more about this.  Maybe the Mathlib API for meromorphic trailing coeff and divisors is going to be too annoying to use for functions which are actually analytic. However you can make your definitions return junk values when the function isn't analytic or the zero set is infinite to avoid lots of extra hypotheses in the definition.  I've been experimenting with this in my fork of the StrongPNT proof, see how I've defined Cf etc in https://github.com/ajirving/strongpnt/blob/main/StrongPNT/PNT2_LogDerivative.lean

prestontranbarger: Okay, I'll give this a look when I get a chance. Thanks again for all of your input!

AlexKontorovich:
For the blueprint tags, we want to copy this style:
```
@[blueprint
  "bklnw-lemma-11b"
  (title := "BKLNW Lemma 11b")
  (statement := /-- With the hypotheses as above, we have
  $$ (1 - \eps(b) - c_0(e^{-b/2} + e^{-2b/3} + e^{-4b/5})) x \leq \theta(x)$$
   for all $x \geq e^b$ and $b>0$, where $c_0 = 1.03883$ is the constant from \cite[Theorem 12]{rs-prime}. -/)
  (proof := /-- From Theorem \ref{costa-pereira-theorem-1a} we have $\psi(x) - \theta(x) ≤ \psi(x^{1/2}) + \psi(x^{1/3}) + \psi(x^{1/5})$.  Now apply the hypothesis on $\psi(x)$ and  Theorem \ref{rs-psi-upper}. -/)
  (latexEnv := "lemma")
  (discussion := 789)]
```

PrimeNumberTheoremAnd/StrongPNT.lean
```diff
@@ -226,36 +226,61 @@ theorem LogOfAnalyticFunction {r R : ℝ} (zero_lt_r : 0 < r) (r_lt_R : r < R)
...
+  (statement := /--
+    Let $f:\mathbb{C}\to\mathbb{C}$.
+    We define $m_f(\rho)$ as the order of the zero $\rho$ w.r.t $f$.
+  -/)]
+noncomputable def ZeroFactor (f : ℂ → ℂ) (z : ℂ) : ℂ :=
+  if h1 : AnalyticAt ℂ f z then
+    if h2 : analyticOrderAt f z ≠ ⊤ then
+      (h1.analyticOrderAt_ne_top.mp h2).choose z
+    else 0
+  else 0
```
AlexKontorovich, line 273: I think [analyticOrderNatAt](https://leanprover-community.github.io/mathlib4_docs/Mathlib/Analysis/Analytic/Order.html#analyticOrderNatAt) already does this?

PrimeNumberTheoremAnd/StrongPNT.lean
```diff
@@ -226,36 +226,61 @@ theorem LogOfAnalyticFunction {r R : ℝ} (zero_lt_r : 0 < r) (r_lt_R : r < R)
...
+
+@[blueprint "ZeroFactorization"
+  (title := "ZeroFactorization")
+  (statement := /--
+    Let $f:\mathbb{C}\to\mathbb{C}$ be analytic on $\overline{\mathbb{D}_1}$ with $f(0)\neq 0$.
+    For all $\rho\in\mathcal{K}_f(R)$ with $R<1$ there exists $h_\rho(z)$ such that
+    $h_\rho(z)$ is analytic at $\rho$, $h_\rho(\rho)\neq 0$, and
+    $f(z)=(z-\rho)^{m_f(\rho)}\,h_\rho(z)$.
+  -/)
+  (proof := /--
```
AlexKontorovich, line 285: do you really need this or can we just quote [analyticOrderAt_ne_top](https://leanprover-community.github.io/mathlib4_docs/Mathlib/Analysis/Analytic/Order.html#AnalyticAt.analyticOrderAt_ne_top) where it's needed?

AlexKontorovich: Actually [MeromorphicNFAt](https://leanprover-community.github.io/mathlib4_docs/Mathlib/Analysis/Meromorphic/NormalForm.html#MeromorphicNFAt) might be even more useful?

PrimeNumberTheoremAnd/StrongPNT.lean
```diff
@@ -346,23 +520,53 @@ blueprint_comment /--
...
-\end{proof}
--/
+  -/)]
+lemma DiskBound {B r R : ℝ} (r_lt_one : r < 1) (R_pos : 0 < R) (r_lt_R : r < R) (R_lt_one : R < 1)
+  {f : ℂ → ℂ} (finiteZeros : (SetOfZeros 1 f).Finite)
+  (hfAnalytic : AnalyticOnNhd ℂ f (Metric.closedBall (0 : ℂ) 1))
+  (hf_neq_zero_at_zero : f 0 ≠ 0) (fz_bound : ∀ (z : ℂ), ‖z‖ ≤ R → ‖f z‖ ≤ B) :
+  ∀ z ∈ Metric.closedBall (0 : ℂ) R,
+    ‖BlaschkeB r R f z‖ ≤ B := by
+  intro z hz
```
AlexKontorovich:
```suggestion
  (hf_neq_zero_at_zero : f 0 ≠ 0) (fz_bound : ∀ (z : ℂ), ‖z‖ ≤ R → ‖f z‖ ≤ B)
  {z : ℂ} (hz : Metric.closedBall (0 : ℂ) R) :
    ‖BlaschkeB r R f z‖ ≤ B := by
```

AlexKontorovich (approved):
Thanks!

Issue #1309, opened 2026-04-27
[MERTENS-MATHLIB]: Mertens' first theorem (von Mangoldt form)


teorth:
## Description

NOTE: As the material in this subproject is intended for Mathlib, the use of autoformalization tools is strongly discouraged, and reliance on other PNT+ results should be avoided if at all possible. Aim to keep proofs short and structured.

Easy combination of previous results

## Blueprint

[Corollary 7.1.4](https://alexkontorovich.github.io/PrimeNumberTheoremAnd/blueprint/explicit-chapter.html#Mertens-first-error-mangoldt-le)

## Zulip

[Link](https://leanprover.zulipchat.com/#narrow/channel/423402-PrimeNumberTheorem.2B/topic/Mathlib-suitable.20version.20of.20Mertens.20theorem.3F)

YanYablonovskiy:
I can have a go if that’s permissible , tend to enjoy the soon to be lost art of hand typing proofs !

teorth:
Sure!  You will need to post the single word "claim" in this issue to claim it. If you also wish to claim previous lemmas then you will need to post there as well.  If you find you need to refactor the proof differently from how it is structured here, you can go ahead and do so (if it is a minor change) or discuss it on the Zulip (if it is a major change). EDIT: Never mind, I just assigned you directly.

ajirving:
This has been closed but E₁Λ.bounded which is tagged with this issue is still a sorry.  Its only a couple of lines so should I just open a PR to fix it?

teorth:
Good catch.  There was another contributor who also patched this recently, but your fix is shorter, so I will implement that one.

PR #1362, opened 2026-05-09
feat(Mertens): prove E₂Λ.eq

### LINKED ISSUE #1317: [MERTENS-MATHLIB]: Expressing second Mertens error in terms of first Mertens error
## Description

NOTE: As the material in this subproject is intended for Mathlib, the use of autoformalization tools is strongly discouraged, and reliance on other PNT+ results should be avoided if at all possible. Aim to keep proofs short and structured.

The given proof requires Riemann-Stieltjes integration which is not currently well supported in Mathlib.  However there should be a way to proceed via the fundamental theorem of calculus and Tonelli's theorem.

## Blueprint

[Corollary 7.1.8](https://alexkontorovich.github.io/PrimeNumberTheoremAnd/blueprint/explicit-chapter.html#Mertens-second-error-mangoldt-eq)

## Zulip

[Link](https://leanprover.zulipchat.com/#narrow/channel/423402-PrimeNumberTheorem.2B/topic/Mathlib-suitable.20version.20of.20Mertens.20theorem.3F)

ajirving:
Proves E₂Λ.eq and the private theorems before it.  Some comments:
* I fixed the definition of gamma which was incorrectly perenthesised.
* I used Abel summation from Mathlib to prove sum_log_div_eq, there was some amount of annoyance translating between different forms of the same sum.
* The proofs are not complete as they depend on #1309 and #1316.  We could remove the latter by not requiring the explicit bound on $E_1$.

Closes #1317


teorth:
Looks good to me.  I'm pretty confident that the previous sorries will be filled in in due course, so this doesn't concern me much.  Thanks for dealing with the integrability issues; I still struggle a bit with the integrability API.

teorth (approved):

PR #1364, opened 2026-05-09
feat(IK): prove zeta_mul_tau_square_eq

### LINKED ISSUE #1244: [IK]: Prove `zeta_mul_tau_square_eq`
## Description

This is the "baby" Rankin-Selberg convolution. Prove `zeta_mul_tau_square_eq`

## Blueprint

https://alexkontorovich.github.io/PrimeNumberTheoremAnd/blueprint/IK-chapter.html#zeta_mul_tau_square_eq

## Zulip

TBA

aeroplugin:
Co-authored with Claude

Closes #1244


PrimeNumberTheoremAnd/IwaniecKowalskiCh1.lean
```diff
@@ -662,6 +662,62 @@ theorem zeta_pow_four_eq (s : ℂ) (hs : 1 < s.re) :
...
+    rw [mul_pow]
+    exact isMultiplicative_tau.2 (hmn.pow 2 2)
+
+lemma isMultiplicative_sqTau : sqTau.IsMultiplicative := by
+  refine ⟨?_, ?_⟩
+  · change τ (1 : ℕ) ^ 2 = 1
+    rw [isMultiplicative_tau.1, one_pow]
+  · intro m n hmn
+    change τ (m * n) ^ 2 = τ m ^ 2 * τ n ^ 2
+    rw [isMultiplicative_tau.2 hmn, mul_pow]
```
AlexKontorovich: I'm realizing that this could use some API as well: any time a multiplicative function is composed (in either order) with a coprimality-preserving function (that is, gcd(n,m) = 1 => gcd(f(n),f(m)) = 1; e.g., x -> x^2), the result is multiplicative. That's the "real" reason for both of these, right?

aeroplugin: Yes, that is right, although it splits into two lemmas, as coprimality preservation by itself isn't quite the right condition for the post-composition case. For sqTau = (·)² ∘ τ: (τ(mn))² = (τ(m)·τ(n))² = τ(m)²·τ(n)² only needs τ multiplicative and (·)² respecting multiplication, where no coprimality preservation is needed on the outer map. For tauSq = τ ∘ (·)²: τ((mn)²) = τ(m²·n²) = τ(m²)·τ(n²), it needs (·)² to both respect multiplication and preserve coprimality, so τ can split. Squaring satisfies both cases, which is why the two cases can be collapsed into a single example. I don't see these in the Mathlib API, specifically in Mathlib.NumberTheory.ArithmeticFunction (.mul, .pmul, .pow are there, but not the monoid-hom version). Should I factor out these two lemmas locally and use them for the Zeta-mul-tau-square-eq proof?

AlexKontorovich: Right, good point. Yes please if you could factor out the two API lemmas, I think that would improve the formalization. Thanks!

aeroplugin:
Thank you for the suggestion! I have factored out the two API lemmas. Please let me know if there is anything else I can do to improve the proof.

PrimeNumberTheoremAnd/IwaniecKowalskiCh1.lean
```diff
@@ -663,6 +663,73 @@ theorem zeta_pow_four_eq (s : ℂ) (hs : 1 < s.re) :
...
+    (hf : f.IsMultiplicative) (g : ℕ →* ℕ) (hh : ∀ n, h n = g (f n)) :
+    h.IsMultiplicative := by
+  refine ⟨?_, ?_⟩
+  · rw [hh, hf.1, g.map_one]
+  · intro m n hmn
+    rw [hh, hh, hh, hf.2 hmn, g.map_mul]
+
+lemma IsMultiplicative.of_comp_mulHom {f h : ArithmeticFunction ℕ}
+    (hf : f.IsMultiplicative) (g : ℕ →* ℕ)
+    (hg_cop : ∀ {m n : ℕ}, Nat.Coprime m n → Nat.Coprime (g m) (g n))
```
AlexKontorovich: Should this hypothesis be a standalone typeclass? Seems like a useful condition actually (and not one I think appears frequently - if at all - in the literature). Something like `IsCoprimePreserving`? You could then instantiate it for the `sqHom` function, and have Lean fill in the class automatically...

PrimeNumberTheoremAnd/IwaniecKowalskiCh1.lean
```diff
@@ -663,6 +663,73 @@ theorem zeta_pow_four_eq (s : ℂ) (hs : 1 < s.re) :
...
+lemma IsMultiplicative.of_map_mulHom {f h : ArithmeticFunction ℕ}
+    (hf : f.IsMultiplicative) (g : ℕ →* ℕ) (hh : ∀ n, h n = g (f n)) :
+    h.IsMultiplicative := by
+  refine ⟨?_, ?_⟩
+  · rw [hh, hf.1, g.map_one]
+  · intro m n hmn
+    rw [hh, hh, hh, hf.2 hmn, g.map_mul]
+
+lemma IsMultiplicative.of_comp_mulHom {f h : ArithmeticFunction ℕ}
+    (hf : f.IsMultiplicative) (g : ℕ →* ℕ)
```
AlexKontorovich, line 696: Can `g` be inferred from `hh`? Or does Lean get confused in practice?

PrimeNumberTheoremAnd/IwaniecKowalskiCh1.lean
```diff
@@ -663,6 +663,73 @@ theorem zeta_pow_four_eq (s : ℂ) (hs : 1 < s.re) :
...
 
+lemma IsMultiplicative.of_map_mulHom {f h : ArithmeticFunction ℕ}
+    (hf : f.IsMultiplicative) (g : ℕ →* ℕ) (hh : ∀ n, h n = g (f n)) :
+    h.IsMultiplicative := by
+  refine ⟨?_, ?_⟩
+  · rw [hh, hf.1, g.map_one]
+  · intro m n hmn
+    rw [hh, hh, hh, hf.2 hmn, g.map_mul]
+
+lemma IsMultiplicative.of_comp_mulHom {f h : ArithmeticFunction ℕ}
```
AlexKontorovich, line 695: I see, you want to keep `h` around because if you just keep it unnamed (as `fun n => f (g n)`) then you have to coerce it to an `ArithmeticFunction`? I guess that makes sense...

PrimeNumberTheoremAnd/IwaniecKowalskiCh1.lean
```diff
@@ -663,6 +663,73 @@ theorem zeta_pow_four_eq (s : ℂ) (hs : 1 < s.re) :
...
+    (hf : f.IsMultiplicative) (g : ℕ →* ℕ)
+    (hg_cop : ∀ {m n : ℕ}, Nat.Coprime m n → Nat.Coprime (g m) (g n))
+    (hh : ∀ n, h n = f (g n)) : h.IsMultiplicative := by
+  refine ⟨?_, ?_⟩
+  · rw [hh, g.map_one, hf.1]
+  · intro m n hmn
+    rw [hh, hh, hh, g.map_mul, hf.2 (hg_cop hmn)]
```
AlexKontorovich:
```suggestion
  refine ⟨by rw [hh, g.map_one, hf.1], fun m n hmn => ?_⟩
  rw [hh, hh, hh, g.map_mul, hf.2 (hg_cop hmn)]
```

PrimeNumberTheoremAnd/IwaniecKowalskiCh1.lean
```diff
@@ -663,6 +663,73 @@ theorem zeta_pow_four_eq (s : ℂ) (hs : 1 < s.re) :
...
+
+abbrev sqHom : ℕ →* ℕ where
```
AlexKontorovich: I guess the squaring function is sufficiently special that it really deserves its own name, right? Should we also have a `powHom` (if it doesn't exist already) sending `n => n ^ k`? And then maybe `sqHom` is a special case `k =2`?

PrimeNumberTheoremAnd/IwaniecKowalskiCh1.lean
```diff
@@ -663,6 +663,73 @@ theorem zeta_pow_four_eq (s : ℂ) (hs : 1 < s.re) :
...
+  toFun n := n ^ 2
+  map_one' := one_pow 2
+  map_mul' m n := mul_pow m n 2
+
+/-- `τ(p^m) = m + 1` for prime `p`. -/
+lemma tau_prime_pow {p : ℕ} (hp : p.Prime) (m : ℕ) : τ (p ^ m) = m + 1 :=
```
AlexKontorovich: Missing `by` at the end?

PrimeNumberTheoremAnd/IwaniecKowalskiCh1.lean
aeroplugin: I am not sure whether lean could infer g from hh. But I can try testing how Lean's higher-order unifier would work.

PrimeNumberTheoremAnd/IwaniecKowalskiCh1.lean
aeroplugin: Great catch, will turn the hypothesis into a standalone typeclass.

PrimeNumberTheoremAnd/IwaniecKowalskiCh1.lean
aeroplugin: Will fix.

PrimeNumberTheoremAnd/IwaniecKowalskiCh1.lean
aeroplugin: I was looking at Mathlib and it already has `powMonoidHom (k : ℕ) : M →* M` 
(in `Mathlib.Algebra.Group.Basic` for commutative monoids). I'll drop 
our local `sqHom` and just use `powMonoidHom 2` directly.

aeroplugin:
Hello, I applied all of your suggestions to my proof of zeta-mul-tau-square-eq. Is there anything else I can do?

PrimeNumberTheoremAnd/IwaniecKowalskiCh1.lean
```diff
@@ -689,9 +759,94 @@ Baby Rankin-Selberg:
...
+    change ((τ n : ℕ) : ℂ) ^ 2 = (((τ n) ^ 2 : ℕ) : ℂ)
+    push_cast
+    rfl
+  rw [lhs_bridge, rhs_bridge, key]
+  
+/--
+Zeta cubed:
+`ζ(s)^3 = ζ(2s) ∑ τ(n^2) n^(-s)`. -/
+@[blueprint
+  "zeta_pow_three_eq"
```
AlexKontorovich:
```suggestion
  "zeta_pow_three_eq'"
```

PrimeNumberTheoremAnd/IwaniecKowalskiCh1.lean
```diff
@@ -689,9 +759,94 @@ Baby Rankin-Selberg:
...
+  (title := "zeta pow three eq")
+  (statement := /-- Zeta cubed: $\zeta(s)^3 = \zeta(2s) \sum_{n=1}^{\infty}\tau(n^2) n^{-s}$.
+  \begin{verbatim}
+  This is IK (1.30).
+  \end{verbatim}
+  -/)
+  (proof := /--
+  This follows from the previous two theorems. From the corollary of Ramanujan's formula, we have $\zeta(s)^4 = \zeta(2s) \sum_{n=1}^{\infty} \tau(n)^2 n^{-s}$. From the Baby Rankin-Selberg result, we have $\zeta(s) \sum_{n=1}^{\infty} \tau(n^2) n^{-s} = \sum_{n=1}^{\infty} \tau(n)^2 n^{-s}$. Combining these two results, we can express $\zeta(s)^4$ in terms of $\zeta(s)$ and $\sum_{n=1}^{\infty} \tau(n^    2) n^{-s}$, which leads to the conclusion that $\zeta(s)^3 = \zeta(2s) \sum_{n=1}^{\infty} \tau(n^2) n^{-s}$.
+  -/)]
+lemma zeta_pow_three_eq (s : ℂ) (hs : 1 < s.re) :
```
AlexKontorovich:
```suggestion
lemma zeta_pow_three_eq' (s : ℂ) (hs : 1 < s.re) :
```

PrimeNumberTheoremAnd/IwaniecKowalskiCh1.lean
```diff
@@ -689,9 +759,94 @@ Baby Rankin-Selberg:
...
+  (title := "zeta pow three eq")
+  (statement := /-- Zeta cubed: $\zeta(s)^3 = \zeta(2s) \sum_{n=1}^{\infty}\tau(n^2) n^{-s}$.
+  \begin{verbatim}
+  This is IK (1.30).
+  \end{verbatim}
+  -/)
+  (proof := /--
+  This follows from the previous two theorems. From the corollary of Ramanujan's formula, we have $\zeta(s)^4 = \zeta(2s) \sum_{n=1}^{\infty} \tau(n)^2 n^{-s}$. From the Baby Rankin-Selberg result, we have $\zeta(s) \sum_{n=1}^{\infty} \tau(n^2) n^{-s} = \sum_{n=1}^{\infty} \tau(n)^2 n^{-s}$. Combining these two results, we can express $\zeta(s)^4$ in terms of $\zeta(s)$ and $\sum_{n=1}^{\infty} \tau(n^    2) n^{-s}$, which leads to the conclusion that $\zeta(s)^3 = \zeta(2s) \sum_{n=1}^{\infty} \tau(n^2) n^{-s}$.
+  -/)]
+lemma zeta_pow_three_eq' (s : ℂ) (hs : 1 < s.re) :
```
AlexKontorovich, line 846:
```suggestion
-- See also `zeta_pow_three_eq` below for a different proof
lemma zeta_pow_three_eq' (s : ℂ) (hs : 1 < s.re) :
```

AlexKontorovich (approved):
Thanks

PR #1365, opened 2026-05-10
feat(Mertens): Euler-Maclaurin and sum_log_eq

### LINKED ISSUE #1303: [MERTENS-MATHLIB]: Partial sum of logarithm identity
## Description

NOTE: As the material in this subproject is intended for Mathlib, the use of autoformalization tools is strongly discouraged, and reliance on other PNT+ results should be avoided if at all possible. Aim to keep proofs short and structured.

A challenge may be that Riemann-Stieltjes integral is not currently well developed in Mathlib, so one may have to find substitutes that are not too unwieldy.

## Blueprint

[Lemma 7.1.1](https://alexkontorovich.github.io/PrimeNumberTheoremAnd/blueprint/explicit-chapter.html#Mertens-sum-log)

## Zulip

[Link](https://leanprover.zulipchat.com/#narrow/channel/423402-PrimeNumberTheorem.2B/topic/Mathlib-suitable.20version.20of.20Mertens.20theorem.3F)

ajirving:
Proves the first order version of Euler-Maclaurin and uses it to prove sum_log_eq.  I've slightly changed the statement of sum_log_eq to integrate $t-[t]-1/2$ rather than $t-[t]$: that seems more natural and gives a smaller error.  Anyway we're not currently using sum_log_eq so hopefully that's fine.

A future task could be to rework ZetaBounds.lean to use this Euler-Maclaurin rather than proving it's own sum_eq_int_deriv.

Closes #1303

### DIFF
--- a/PrimeNumberTheoremAnd/Mertens.lean
+++ b/PrimeNumberTheoremAnd/Mertens.lean
@@ -27,6 +27,77 @@ theorem one_eq_o_log_log : (fun _ ↦ (1:ℝ)) =o[atTop] (fun x ↦ log (log x))
 
 end Real
 
+section EulerMaclaurin
+open Finset Interval MeasureTheory
+
+variable {𝕜 : Type*} [RCLike 𝕜] {f : ℝ → 𝕜} {a b : ℝ}
+
+noncomputable def B1 (x : ℝ) : ℝ := x - ⌊x⌋₊ - 1 / 2
+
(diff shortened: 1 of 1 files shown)


teorth:
Great!  I need to start upstreaming bits and pieces of this file into Mathlib, now that I have some experience in doing so.

teorth (approved):

PR #1390, opened 2026-05-17
Chore: update to V4.29.0

ajirving:
Updating to V4.29.0, requires numerous fixes:
* Many theorems need ``backward.isDefEq.respectTransparency`` or else they fail for mysterious reasons.  Results involving integrals especially integral_const_mul are particularly sensative to this.
* A Mathlib change to the definition of distance in commutative groups means that various expressions are no longer definitionally equal so we have to use ``dist_eq_norm_sub``.
* Various places were using norm_num to simplify derivative calculations.  I couldn't get these to work with either norm_num or simp so had to explicitly rewrite with ``deriv_fun_add`` etc.
* Some changes to the output of ring_nf broke some delicate proofs.
* Replaced ``push_neg`` in many places.


AlexKontorovich (approved):
Thank you!!

PR #1419, opened 2026-05-24
fix: close unique_divisor_decomposition and sum_divisors_mul_of_coprime sorrys

d0d1:
## Summary

Closes two sorrys in `IwaniecKowalskiCh1.lean` that were marked `UPSTREAMED TO MATHLIB #36495`. Mathlib PR #36495 is not merged at our pinned Mathlib (8a178386ff, v4.29.0), so we provide direct proofs using the existing Mathlib API.

### `unique_divisor_decomposition` (L75)
For coprime `a`, `b` and `d ∣ a * b`, the unique pair is `(d.gcd a, d.gcd b)`.
- Existence via `Nat.gcd_mul_gcd_eq_iff_dvd_mul_of_coprime`
- Uniqueness via `Nat.Coprime.of_dvd_left/right` + `Nat.dvd_antisymm`

### `sum_divisors_mul_of_coprime` (L90)
2-line proof:
```lean
simp only [← coe_mul_zeta_apply]
exact (hf.mul isMultiplicative_zeta.natCast).map_mul_of_coprime hab
```
Rewrites sums-over-divisors to Dirichlet convolution with ζ, then applies `IsMultiplicative.map_mul_of_coprime` to the product `f * ζ`.

## Axiom verification

Both theorems: `[propext, Classical.choice, Quot.sound]` — no `sorryAx`.

### DIFF
--- a/PrimeNumberTheoremAnd/IwaniecKowalskiCh1.lean
+++ b/PrimeNumberTheoremAnd/IwaniecKowalskiCh1.lean
@@ -72,7 +72,26 @@ lemma IsCompletelyAdditive.isAdditive [AddZeroClass R] {f : ArithmeticFunction R
   -/)]
 lemma unique_divisor_decomposition {a b d : ℕ} (hab : Coprime a b) (hd : d ∣ a * b) :
     ∃! p : ℕ × ℕ, p.1 ∣ a ∧ p.2 ∣ b ∧ p.1 * p.2 = d := by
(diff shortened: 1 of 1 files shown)


AlexKontorovich:
In the IK project, you should only work on sorry's that are in the project board, and only those that are unclaimed there. This `sorry` is not one that we want to close here; it's been PRed to Mathlib, so once it gets in, this theorem is intended to be removed. So I'm going to close this PR. (And likely others, for similar reasons)

PR #1426, opened 2026-05-25
Prove bklnw_table_from_buthe (#1261)

### LINKED ISSUE #1261: [BKLNW]: Importing Table 1 from Buthe
## Description

Should be straightforward (but only a small number of rows from Buthe's table are actually used here)

## Blueprint

[Lemma 10.8.6](https://alexkontorovich.github.io/PrimeNumberTheoremAnd/blueprint/secondary-chapter.html#bklnw-table_from_buthe)

## Zulip

[Link](https://leanprover.zulipchat.com/#narrow/channel/423402-PrimeNumberTheorem.2B/topic/Starting.20on.20BKLNW)

gotrevor:
Draft for review. Does **not** discharge a net sorry — instead, it splits the original sorry into (a) the real combinatorial proof, fully done here, and (b) a single named stub `Buthe.sieve_bound` representing Buthe's Eq. (6.2) sieve verification. Wanted to surface this trade-off explicitly before claiming the issue.

Closes #1261 contingent on `Buthe.sieve_bound`.

## What this PR does

Proves `BKLNW.bklnw_table_from_buthe` by dispatching over the 3 tuples in `table_from_buthe`, and for each tuple doing a sub-interval case analysis combining:

- **Buthe Eq. (6.2)** on `[100, 5×10^10]` — added as a new lemma `Buthe.sieve_bound` (axiomatized, see below)
- The relevant rows of **Buthe Table 1** on `[5×10^10, v]`, via the existing `Buthe.eq_6_2`

Per-tuple coverage:

| Tuple | `(u, v, c, C)` | Sub-intervals |
|---|---|---|
| 1 | `(100, 5×10^10, 0.8, 0.81)` | sieve only — `[-0.8, 0.81]` is *exactly* Eq. (6.2) |
| 2 | `(100, 32×10^12, 0.88, 0.86)` | sieve + 10 Table 1 rows; `max\|M⁻\|=0.88`, `max M⁺=0.86` |
| 3 | `(100, 10^19, 0.94, 0.94)` | sieve + 29 Table 1 rows; `max\|M⁻\|=0.94` (rows 27,28), `max M⁺=0.94` (row 30) |

Adds two private helpers `apply_buthe_row` / `apply_buthe_sieve` that relax a per-row `[Mψ⁻, Mψ⁺]` interval to the looser per-tuple `[-c, C]`.

## On the new `Buthe.sieve_bound` stub

Buthe 2018 (Math. Comp. 87) states Eq. (6.2) directly:

> *In addition the bounds* `−0.8 ≤ R_ψ(t) ≤ 0.81` *for* `100 ≤ t ≤ 5 × 10^10` *have been calculated using the Eratosthenes sieve.*

The existing `Buthe.eq_6_2` lemma in the project, despite its name, actually states Table 1 (the per-doubling-interval bounds), not Eq. (6.2). The literal `Eq. (6.2)` sieve bound is needed for Tuple 1 (where `(c, C) = (0.8, 0.81)` is *tighter* than what any Table 1 row gives) and for the low end of Tuples 2 & 3.

Rather than inline the sieve range as ad-hoc bounds inside `bklnw_table_from_buthe`, this PR adds a named `Buthe.sieve_bound` lemma that mirrors Buthe's paper structure. It's left as `sorry` — the underlying computation would be either a certified sieve (analogous to how the project handles other Buthe paper-claims like `theorem_2a` through `theorem_2f`, which are all currently `sorry`) or imported as an axiom. Happy to file a separate follow-up issue for this if maintainers prefer.

## Build verification

- `lake build` succeeds on the whole project, no regressions
- `#print axioms BKLNW.bklnw_table_from_buthe` shows the expected `[propext, sorryAx, Classical.choice, Quot.sound]` — `sorryAx` is transitively from `Buthe.eq_6_2` (pre-existing) and `Buthe.sieve_bound` (added here)
- The lemma uses `set_option maxHeartbeats 1200000` due to the 30-way case split's elaboration cost; a leading comment documents this

## Caveats / things I'd appreciate guidance on

1. **Process**: I didn't comment `claim` on #1261 before working. Opening as **draft** to verify the approach before claiming. If the direction looks good I'll claim and convert to ready-for-review.
2. **The sorry trade-off** described above — single-PR with both pieces vs. splitting into two PRs (sieve_bound first, then this proof on top).
3. **Naming of `Buthe.eq_6_2`** — its current label points to Eq. (6.2) but its statement is Table 1. Did not rename in this PR but flagging it.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/PrimeNumberTheoremAnd/BKLNW.lean
+++ b/PrimeNumberTheoremAnd/BKLNW.lean
@@ -1597,6 +1597,28 @@ def table_from_buthe : List (ℝ × ℝ × ℝ × ℝ) := [
   (100, 10 ^ 19, 0.94, 0.94)
 ]
 
+/-- Apply a row of `Buthe.table_1` to bound `R_ψ(t) = (t - ψ t) / √t` on `[x, 2x]` and
+relax the per-row `[Mψ⁻, Mψ⁺]` interval to a looser `[-c, C]`. -/
+private lemma apply_buthe_row {x Mψ_minus Mψ_plus t c C : ℝ}
+    (h_row : (x, Mψ_minus, Mψ_plus) ∈ Buthe.table_1)
+    (h_lb : -c ≤ Mψ_minus) (h_ub : Mψ_plus ≤ C)
+    (h_t : t ∈ Set.Icc x (2 * x)) :
+    -c ≤ (t - ψ t) / sqrt t ∧ (t - ψ t) / sqrt t ≤ C := by
+  obtain ⟨h1, h2⟩ := Buthe.eq_6_2 x Mψ_minus Mψ_plus h_row t h_t
+  exact ⟨h_lb.trans h1, h2.trans h_ub⟩
+
+/-- Apply Buthe's sieve bound (Eq. 6.2) on `[100, 5×10^10]` and relax to `[-c, C]`. -/
+private lemma apply_buthe_sieve {t c C : ℝ}
+    (h_lb : -c ≤ (-0.8 : ℝ)) (h_ub : (0.81 : ℝ) ≤ C)
+    (ht1 : 100 ≤ t) (ht2 : t ≤ 5 * 10 ^ 10) :
+    -c ≤ (t - ψ t) / sqrt t ∧ (t - ψ t) / sqrt t ≤ C := by
+  obtain ⟨h1, h2⟩ := Buthe.sieve_bound t ht1 ht2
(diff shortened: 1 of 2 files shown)


gotrevor:
Closing — opened publicly by mistake, didn't intend to surface yet. Will reopen when ready.

teorth:
Note that @Robby955 has now addressed this particular issue.

PR #1467, opened 2026-06-01

Chessing234:
## Summary

`LSeries_totient_eq` stated the Euler product identity with hypothesis `1 < Re(s)`, but the Dirichlet series is only absolutely summable for `Re(s) > 2`.

## Root cause

The convergence region was copied from the IK identity without matching the absolute-summability threshold.

## Why this fix is correct

For `Re(s) ∈ (1, 2]`, terms are bounded by `n^{1-Re(s)}`, which is not absolutely summable. The identity requires `2 < s.re`, matching the standard convergence region for this L-series.

Complements #1468 (blueprint text).

## Test plan


Chessing234:
Rebased onto latest main; the Lean hypothesis is now  to match absolute convergence.

Chessing234:
awaiting-review

Chessing234:
Superseding with a rebased PR.

PR #1471, opened 2026-06-02
[StrongPNT] delete DerivativeBound.lean

prestontranbarger:
I'm fairly certain this file is unnecessary and we have a proof in StrongPNT.lean itself.

Additionally, continuing to keep it is causing some problems. See that Theorem 5.3.2 on the blueprint has no checkmark because there are two theorems named DerivativeBound (up to case), and one of them (this one) still has a sorry.


AlexKontorovich (approved):
Thanks!

PR #1472, opened 2026-06-03
chore: migrate to Lean 4.30.0

alerad:
## Summary

Migrates PNT+ to Lean 4.30.0.

- Bump `lean-toolchain` to `leanprover/lean4:v4.30.0`
- Bump Mathlib, LeanCert, PrimeCert, and related Lake dependencies to 4.30-compatible revisions
- Remove local Chebyshev lemmas that are now provided upstream by Mathlib
- Update proof scripts for Lean/Mathlib 4.30 drift
- Update the local `additive_combination` import path for Mathlib’s renamed linear-combination module

## Verification

```bash
lake update
lake build PrimeNumberTheoremAnd

### CODE REVIEW

File: PrimeNumberTheoremAnd/IEANTN/Mertens.lean [hunk 1]
```diff
@@ -677,7 +677,14 @@ private theorem integrable_E₁Λ_div_mul_log_sq {x : ℝ} (hx : 2 ≤ x) :
     MeasureTheory.IntegrableOn (fun x ↦ E₁Λ x / (x * log x ^ 2)) (Set.Ioi x) MeasureTheory.volume := by
   obtain ⟨c, hc1, hc2⟩ := E₁Λ.bounded'
   apply MeasureTheory.Integrable.mono (integrable_const_div_mul_log_sq c hx)
-  · exact Measurable.aestronglyMeasurable (by fun_prop)
+  · have hsum :
```
ajirving: You can get the original fun_prop proof to work by putting ``attribute [fun_prop] measurable_from_top`` somewhere before this declaration.  That also fixes the similar issue further down the file.  Not sure why Mathlib dropped fun_prop on that theorem.

ajirving:
Thanks for doing this.  You should revert my change from #1414 to use the main version of docgen-action because the issue we had has been fixed with V4.30.

lakefile.toml
```diff
@@ -15,12 +15,12 @@ linter.flexible = true
 [[require]]
 name = "LeanArchitect"
 git = "https://github.com/hanwenzhu/LeanArchitect.git"
-rev = "v4.29.0"
+rev = "v4.30.0-rc2"
```
ajirving: Should this not be V4.30.0?

alerad: My bad yeah, just pushed with v4.30 and added `attribute [fun_prop] measurable_from_top`

ajirving:
Please can you fix any deprecation warnings that remain?  There are also lots of warnings about the copywright header, maybe we can just disable the header linter for now.

ajirving:
Thank you, this looks good now.

teorth (approved):
Thanks for all your work on this!  Interesting that such tasks still have to be done mainly by hand - of all the tasks susceptible to automation, I would have thought Mathlib bumping to be near the top.  Anyway, looks good to me!

Issue #1255, opened 2026-04-05
[BKLNW]: Verification of Table 10


teorth:
## Description

A large amount of numerical computation

## Blueprint

[Proposition 10.8.4](https://alexkontorovich.github.io/PrimeNumberTheoremAnd/blueprint/secondary-chapter.html#bklnw-table-10-verification)

## Zulip

[Link](https://leanprover.zulipchat.com/#narrow/channel/423402-PrimeNumberTheorem.2B/topic/Starting.20on.20BKLNW)

Robby955:
I looked into this locally and found what appears to be a statement mismatch rather than a row-verification/tactic issue.

For the first row of Table 10, `b = 20`, `table_10_next 20 = 21`, and for `k = 5` the table value is `2.9160e2 = 291.60`.

The current Lean target asks for

```lean
B_8_1 5 20 (table_10_next 20) <= 291.60
```

But `B_8_1` appears to be the endpoint upper bound corresponding to the weaker `\widetilde B_k` in Lemma 8 / Corollary 8.1, while Table 10 appears to list the exact interval maximum `B_k(b,b')`.

In my local scratch file, the data-level obstruction is already visible:

```lean
291.60 < 20^5 * exp(-20/2) + 21^5 * 4.2676e-5
```

so the endpoint expression is too large for the published Table 10 value. Numerically, for this row:

- Table 10: `291.60`
- exact interval max: about `291.5985`
- endpoint `B_8_1`: about `326.965`

Does the project intend `bklnw_table_10_verification` to verify the exact `B` from Lemma 8 rather than the endpoint bound `B_8_1`?

I have not opened a PR because this seems like a statement/design question first.

teorth:
Ah, I think you're right.  The published reference is a bit confusing because the quantity B_k(b,b') defined on equation (3.13) of Corolllary 8.1 is not compatible with the quantity of the same name defined by equation (3.10) in Lemma 8.  To add to the confusion, the caption for Table 10 refers to Corollary 8.1 instead of Lemma 8.  But given your numerics it seems that one should indeed work with the `B` quantity from Lemma 8 rather than the `B_8_1` quantity from Corollary 8.1.  Feel free to make a PR accordingly!

Robby955:
Thanks — that matches what I'm seeing, and I've started working through the PR shape locally.

[...]

PR #1510, opened 2026-06-09
Pin the Table-10 b=20,k=5 statement mismatch (#1255): exact Lemma-8 B vs Cor-8.1 endpoint

Robby955:
Picking up #1255. Two small compiled facts for the first row of Table 10 (b=20, k=5) that pin down the statement mismatch you confirmed in the thread, plus a soundness cleanup for the a₂ cert they rely on.

**1. The current `bklnw_table_10_verification` target is false at b=20.** `table10_verification_statement_false` proves `¬(…)` for *exactly* the statement of the live `sorry`'d `bklnw_table_10_verification` — the target that bounds the Corollary-8.1 endpoint quantity `B_8_1`. At b=20, `B_8_1 5 20 21 ≈ 326.96 > 291.60` (the listed value), so the universal is provably false. The numeric core is `sorryAx`-free.

**2. The k=5 exact Lemma-8 B satisfies the listed k=5 bound.** `table_10_row20_k5 : B_8_exact 5 20 21 ≤ 291.60`, where `B_8_exact 5 20 21` is *definitionally* the Lemma-8 interval supremum `B 5 2 a Inputs.default.ε 20 21` (the iSup used inside `bklnw_cor_8_1a`; checked by `rfl`). The supremum is ≈ 291.5985, so the 291.60 bound is tight at the displayed two-decimal precision, not a loose margin. The reduction uses convexity to reduce the supremum to endpoint checks, then `interval_decide` closes the comparison (the right endpoint gives the maximum); that numeric core is `sorryAx`-free.

Together these formalize the b=20, k=5 instance of the (3.13)-vs-(3.10) reading from the thread: this displayed Table-10 entry matches the Lemma-8 `B`, not the Corollary-8.1 endpoint `B_8_1`.

**Scope (one row, one column).** This PR's *theorems* discharge only the k=5 anchor at b=20; I am *not* claiming row 20 verifies at the listed values across all columns. (Auxiliary computation, not formalized in this PR: for the full row, k=1,2 only pass with the repo's table margin `table_8_margin·1.001`, because the bare 5-figure entries *undershoot* the exact-B sup by ~1.9e-8 and ~3.8e-7 respectively; k=3,4,5 hold at the bare value.) The k=5 anchor is the cleanest single defensible statement and exhibits the (3.13)-vs-(3.10) gap most sharply (326.96 vs 291.60).

**Honest status / axioms.** `#print axioms` on both theorems lists a `sorryAx`. It enters *only* through `Inputs.default`, whose proof fields `hε`, `hx₁'`, `hα` reduce to `BKLNW_app.theorem_2`, `buthe_eq_1_7`/`Buthe.theorem_2c`, and `cor_2_1` (the ψ−x and small-x θ analytic inputs), all still on `sorry` upstream. **No new `sorry` is introduced**, and the `sorryAx` is not load-bearing for the two numeric inequalities. So this is "correct *modulo* the in-progress upstream analytic inputs," not a standalone axiom-clean proof.

**a₂ cert cleanup (separable — first commit).** The row-20 a₂ bound goes through `a2_20_mem_Icc`, which delegated to `LeanCert.Examples.BKLNW_a2_bounds.a2_20_exp_*` (`:= by sorry` stubs). The sound `native_decide` equivalents already exist in `LeanCert.Examples.BKLNW_a2_reflective` with identical statements; re-pointing the a2_20 delegation makes `a2_20_mem_Icc` `sorryAx`-free. Happy to drop or split this commit if you'd prefer.

**On process (in the spirit of openness about AI use).** This was developed with an AI assistant under my direction, with every claim gated on the Lean compiler and `#print axioms`. The honest lesson was how much the *verification* did the work — an axiom trace caught a reused `a₂` certificate that was silently `sorry`-backed (now re-routed to the sound `native_decide` proof), and a cross-model review caught a sign error in my own margin wording — i.e. the leverage was in catching the assistant's mistakes, not in trusting its output.

---

**Design question (the part that's yours to decide).** The all-rows `bklnw_table_10_verification` as stated is *unprovable even after the `B_8_1` → exact-`B` retarget*, because some printed rows are wide-gap on the coarse displayed `table_10` grid (e.g. `table_10_next 56 = 2275` rather than the next fine-grid entry 57), so `next(b)` overshoots there. The unabridged Table 10 (anc PDF) has the fine grid where every strip is tight. How would you prefer to handle this — (a) replace `table_10` with the unabridged grid so `next(·)` is per-strip, (b) keep the coarse grid but retarget to the exact `B`, or (c) a separate fine-grid table for the per-strip verification while the coarse grid stays for the cumulative `cor_8_1b`? I have row templates compiled for all three regimes and both special cases (the `if_neg` a₁ branch and the `19·log 10` row) and can follow whichever convention you pick.

### DIFF
--- a/PrimeNumberTheoremAnd/IEANTN/BKLNW/BKLNW_a2_bounds.lean
+++ b/PrimeNumberTheoremAnd/IEANTN/BKLNW/BKLNW_a2_bounds.lean
@@ -6,6 +6,7 @@ Authors: LeanCert Contributors
 import LeanCert.Tactic.IntervalAuto
 import LeanCert.Examples.BKLNW_a2_pow2
 import LeanCert.Examples.BKLNW_a2_bounds
+import LeanCert.Examples.BKLNW_a2_reflective
 import PrimeNumberTheoremAnd.IEANTN.BKLNW.BKLNW
 
 /-!
@@ -47,13 +48,13 @@ private lemma floor_20 : ⌊(20 : ℝ) / log 2⌋₊ = 28 := by
 private lemma a2_20_exp_lower :
     (1.4262 : ℝ) ≤ (1 + 193571378 / (10 : ℝ) ^ 16) * f (exp (20 : ℝ)) := by
   simpa [f_eq_leancert_f] using
-    LeanCert.Examples.BKLNW_a2_bounds.a2_20_exp_lower
+    LeanCert.Examples.BKLNW_a2_reflective.a2_20_exp_lower
 
 private lemma a2_20_exp_upper :
     (1 + 193571378 / (10 : ℝ) ^ 16) * f (exp (20 : ℝ)) ≤
       (1.4262 : ℝ) + (1 : ℝ) / 10 ^ 4 := by
   simpa [f_eq_leancert_f] using
-    LeanCert.Examples.BKLNW_a2_bounds.a2_20_exp_upper
+    LeanCert.Examples.BKLNW_a2_reflective.a2_20_exp_upper
 
 private lemma cert_pow29_upper :
     (1 + 193571378 / (10 : ℝ) ^ 16) * f ((2 : ℝ) ^ (29 : ℕ)) ≤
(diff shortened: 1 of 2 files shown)


teorth:
It seems the cleanest thing to do is replace Table 10 with the unabridged Table 10 throughout the Lean formalization.

PR #1511, opened 2026-06-09
BKLNW: use unabridged Table 10 and retarget verification to Lemma-8 B

Robby955:
This PR adresses the data/specification part of #1255.  The per-row verification proof is still open. This PR replaces the abridged Table 10 data with the unabridged computational grid and retargets `bklnw_table_10_verification` to the exact Lemma 8 quantity tracked by that table.

The previous statement used Corollary 8.1 endpoint quantity `B_8_1`. That target was too large for at least the first sharp row (`b = 20`, `k = 5`), while the table values track the exact interval quantity now exposed as `B_8_exact`.

## Summary

- Replaces `table_10` with the unabridged 287-row grid.
- Leaves `table_10_next` unchanged; the denser table data makes successor lookups such as `table_10_next 56 = 57` definitional.
- Retargets `bklnw_table_10_verification` from `B_8_1` to `B_8_exact`.
- Keeps `cor_8_1a` in its existing `B_8_1` form for compatibility with `cor_8_1b`.
- Exposes the exact-B intermediate and adds application corollaries bridging the exact-B and `B_8_1` forms.

Edited: Removed AI internal process notes from description.


teorth:
If one is to switch from B_8_1 to B_8_exact in bklnw_table_10_verification then a similar modification may need to be made to  bklnw_cor_8_1a, otherwise this theorem will not be usable in applications.

By the way, I would greatly prefer interacting with a human summarizing their AI output, than directly having to respond to unprocessed AI output.

Robby955:
> If one is to switch from B_8_1 to B_8_exact in bklnw_table_10_verification then a similar modification may need to be made to bklnw_cor_8_1a, otherwise this theorem will not be usable in applications.
> 
> By the way, I would greatly prefer interacting with a human summarizing their AI output, than directly having to respond to unprocessed AI output.

Absolutely, thank you for the advice. I will personally take my time to review these questions in careful detail. I wrongly thought it would make it more structured or easier to review, that is on me.

Rob,

Robby955:
Regarding the chaining issue, I have updated the PR to hopefully address that.

The exact bound inside `cor_8_1a` appears as an intermediate step, just before weakening `B` to `B_8_1`. I exposed that step as a companion lemma, `bklnw_cor_8_1a_exact`, and I checked that the table verification chains through to the desired application bound with the listed constant.

I did not change `cor_8_1a`, since `cor_8_1b` currently relies on the `B_8_1` version. Could that be a right approach, or would it be better for the exact-`B` version to become the primary statement instead?

Thanks,

Rob

teorth:
In general it makes sense to expose intermediate lemmas and have multiple variants of the main results to maximize the chance that one of them can be usable for downstream applications; this is certainly in line with the Mathlib philosophy.  So if for instance one can have one version of `cor_8_1a` using `B_8_1` and then some quick corollaries using other versions (or vice versa), then we should have the best of both worlds.

Robby955:
Thanks, I added the “best of both worlds” version.

`cor_8_1a` remains in `B_8_1` for compatibility with `cor_8_1b`. I also exposed the exact-B intermediate and added the application corollaries, with a bridge from `B_8_exact` to `B_8_1`, so downstream code can use either form.

teorth:
Superseded by #1517, so am now closing.

PR #1405, opened 2026-05-21
feat(BKLNW): prove bklnw_table_12_verification (#1263)

Vilin97:
# From me:
Claude did this in 67 minutes from prompt "pick the easiest of tasks among #1255, #1257, #1263 and complete it". It seems that it found a few potential errors in the statements. Proposes #1263.

# From AI:
Verifies Table 12 of [BKLNW](https://arxiv.org/abs/2002.11068).

Four block-boundary rows of `table_12` -- b = log(5e10), 25, log(3.2e13), 32 -- had `Cb` entries computed with the previous block's Buthe constants (c, C), an erratum in the paper that made the verification false as stated. These rows are recomputed from the `C_bk` formula with each row's own (c, C); row b = 31 is raised in columns k = 3,4,5 so the column stays pointwise above the corrected log(3.2e13) row, as `table_12_Cb_bounds` requires.

The numerical check `table_12_check` is added to `BKLNW_tables.lean`, discharging each row with LeanCert's `interval_decide`; the two b = log(N) rows go through a monotonicity helper `C_bk_log_row_bound` because `interval_decide` cannot reduce `exp(-log N / k)` directly.

PR #1517, opened 2026-06-10
BKLNW: prove bklnw_table_10_verification over the unabridged Table 10

Robby955:
This proves `bklnw_table_10_verification` over the full unabridged 287-row Table 10, all five columns, with no new sorries; the remaining axiom profile is inherited from the in-progress upstream results (`theorem_2`, `Buthe.theorem_2c`). It includes the changes currently open in #1511, so either that merges first and I rebase this, or this one can absorb it.

The margin is `table_10_margin := table_8_margin * 1.001`, the same convention as the merged Table 14 (#1405). While porting the grid I found one transcription error: the `table_8` entry at b = 13500 had exponent -34 instead of -35 (the source prints 7.21540e-35, line 2574 of the Full Tables file). That fix is the first commit. The verification theorem moved to `BKLNW_table10_dispatch.lean` because the row certificates it consumes import `BKLNW.lean`; the statement and blueprint attribute are unchanged. A clean build adds about 15 minutes locally; I can split this into smaller PRs if that is easier on CI.

One question: should the margin sit in the statement (`≤ B k * table_10_margin`, as now) or be folded into the stored values so the theorem reads `≤ B k`? Whichever you prefer. (margin is in the statement as written (≤ B k * table_10_margin)) so this remains unchanged.

Tool Usage:

  Lean was developed using Claude Code CLI (Opus 4.8, thinking mode extra-high), plus Python scripts that generate the row 
 certificates and check every literal against the published tables before emission.
 
Drafts and proof strategies also got review passes from manually using GPT-5.5 Pro Extended. Everything is kernel-checked, and I reviewed and assembled the final branch. Builds ran locally on my Apple M5 Pro, 64 GB.


teorth:
I'd put the margin in the statement to emphasise its presence.

By the way, can you record what AI tools were used in this PR?

Robby955:
Of course. I am more than happy to share in whichever format or level of detail you prefer.

I will update with full details on the tools soon, but to be short I use Claude Code CLI and Apple M5 Pro 64 GB.

teorth (approved):

teorth:
As discussed in https://leanprover.zulipchat.com/#narrow/channel/423402-PrimeNumberTheorem.2B/topic/Starting.20on.20BKLNW/with/602845105 , it seems that this PR caused a large increase in build time.  If you have some suggestions on how the build time could be lowered, that would be appreciated.  (Sorry, I should have flagged this earlier; in particular, the lemmas which required increased heartbeats will likely need to be refactored.)

teorth:
Claude is making some headway in reducing the build time.  A PR will be incoming soon.

PR #1526, opened 2026-06-11

Chessing234:
## Summary
`LSeries_totient_eq` used hypothesis `1 < Re(s)`, but `LSeries (↗totient) s` is only absolutely summable for `Re(s) > 2`. In the strip `1 < Re(s) ≤ 2` the series does not converge absolutely, so the identity is false as stated.

## Why this fix is correct
The Euler product for φ requires `Re(s) > 2` and `Re(s-1) > 1`, i.e. `Re(s) > 2`, matching IK (1.35).

## Test plan


Chessing234:
Superseded by #1626 on current main.

PR #1530, opened 2026-06-12
put in sublemmas for proof of Theorem 3.1

teorth:


Robby955:
Reading through the sublemma chain, I may be missing something, but I think there could be a multiplicity issue around eq_12.

The contour integral picks up residues of -ζ'/ζ at zeros, and a zero of order m contributes residue m. However, eq_12 appears to sum over the zero set without an explicit multiplicity weight, while the final statement is written using the multiplicity-weighted zeroes_sum. As written, I only see the composition immediately if the zeros inside the contour are simple.

Robby955:
Another one from the same read-through, on functional_eq. I may be misreading a convention, but the sign of the ζ'/ζ(1-s) term looks flipped: differentiating Λ(s) = Λ(1-s) gives -ζ'/ζ(s) = -log π + ζ'/ζ(1-s) + (1/2)(ψ(s/2) + ψ((1-s)/2)), with a plus on the ζ'/ζ(1-s) term where the current statement has a minus. As a numerical check at s = 2, the stated right side evaluates to about -3.40 while -ζ'/ζ(2) is about 0.57.

A smaller note on the same node: s = -2 satisfies the current hypotheses while ζ(-2) = 0 makes the left side degenerate, so the exclusion may want the trivial zeros as well. the plus-sign version does go through once the zero exclusions are widened to all zeros of ζ at s and 1−s (the digamma terms need the trivial ones excluded anyway)

teorth:
Good catch. It appears there was a sign error in the original source paper, but fortunately it does not seem to impact downstream theorems.

Robby955:
Following up with a numerical check on the sign cascade, since the statement merged in #1529 was transcribed from the same display and I wanted to know which side of it was wrong. With a two-sided test function the minus form balances to about 3e-25, so the corrected sign on this branch looks right, and the plus form in the merged statement inherits the paper's error; correcting it goes with this branch once the Φ definition is two-sided.

One related observation from the same computation: the Φ in the I-definitions is one-sided as written (the integral is over Ioi 0), and with that reading the inversion picks up the jump correction instead of the reflected sum: numerically the residual with no reflected term equals (1/2)φ(0)·log π to about 3e-10. Defining Φ as the two-sided integral makes eq_13 and eq_14 come out as intended.

Issue #1011, opened 2026-02-18
[IK]: prove `sum_divisors_mul_of_coprime`


AlexKontorovich:
## Description

Prove `sum_divisors_mul_of_coprime`

## Blueprint

[Link](https://alexkontorovich.github.io/PrimeNumberTheoremAnd/blueprint/IK-chapter.html#sum_divisors_mul_of_coprime)

## Zulip

TBD

zw810-ctrl:
theorem sum_divisors_mul_of_coprime {R : Type*} [CommRing R]
    {f : ArithmeticFunction R} (hf : f.IsMultiplicative)
    {a b : ℕ} (hab : Coprime a b) (ha : a ≠ 0) (hb : b ≠ 0) :
    ∑ d ∈ (a * b).divisors, f d = (∑ d ∈ a.divisors, f d) * (∑ d ∈ b.divisors, f d) := by
  let g : ℕ × ℕ → ℕ := fun p ↦ p.1 * p.2

  -- (ab).divisors = Image
  have h_image : (a * b).divisors = (a.divisors ×ˢ b.divisors).image g := by
    ext d
    constructor
    · intro hd
      obtain ⟨p, ⟨hp1, hp2, hp_eq⟩, _⟩ := unique_divisor_decomposition hab (mem_divisors.mp hd).1
      rw [mem_image]
      use p
      constructor
      · rw [mem_product, mem_divisors, mem_divisors]
        exact ⟨⟨hp1, ha⟩, ⟨hp2, hb⟩⟩
      · exact hp_eq
    · intro hd
      rw [mem_image] at hd
      obtain ⟨p, hp, rfl⟩ := hd
      rw [mem_product, mem_divisors, mem_divisors] at hp
      rw [mem_divisors]
      exact ⟨mul_dvd_mul hp.1.1 hp.2.1, mul_ne_zero hp.1.2 hp.2.2⟩

  -- Injectivity
  have h_inj : Set.InjOn g (↑(a.divisors ×ˢ b.divisors)) := by
    intro p1 hp1 p2 hp2 h_eq
    simp only [Finset.mem_coe, Finset.mem_product] at hp1 hp2
    dsimp [g] at h_eq

    have h_dvd1 : p1.1 ∣ a := (Nat.mem_divisors.mp hp1.1).1
    have h_dvd2 : p1.2 ∣ b := (Nat.mem_divisors.mp hp1.2).1
[...]

PR #1554, opened 2026-06-13
IK: fill three divisor-sum infra lemmas (#1011)

Robby955:
## Motivation

Fills two of the divisor-sum infrastructure sorries in `PrimeNumberTheoremAnd/IwaniecKowalskiCh1.lean`. Refs #1011. Both close by reuse of existing Mathlib API; no new helpers, no new declarations.

## Scope

- `sum_divisors_mul_of_coprime`: for a multiplicative `f` and coprime nonzero `a`, `b`, the sum of `f` over the divisors of `a * b` factorises as the product of the sums over the divisors of `a` and of `b`. Proved by applying `IsMultiplicative.map_mul_of_coprime` to `ζ * f`. The proof does not need the nonzeroness hypotheses, so their binders are renamed `_ha`, `_hb` to keep the build warning-free; the statement is otherwise unchanged.
- `divisors_mul_injective`: the product map `(d, e) ↦ d * e` is injective on `m.divisors ×ˢ n.divisors` for coprime `m`, `n`; a short `dvd_antisymm` argument.
- History: this PR originally also filled `Nat.sum_divisorsAntidiagonal_prime_pow`; that landed on main via #1740 and the chunk was dropped here. `unique_divisor_decomposition` is left alone since mathlib #36495 adds the corresponding API.
- `@[blueprint]` metadata untouched. One file changed.

## Test plan

- [x] `lake build PrimeNumberTheoremAnd.IwaniecKowalskiCh1`: clean on the branch rebased onto current main.
- [x] `lake build :blueprint`: clean.
- [x] `#print axioms` on both lemmas: `[propext, Classical.choice, Quot.sound]`. No `sorryAx`.
- [x] `sorry` count in the file goes 7 to 5. The diff adds none and removes two.
- [x] No new warnings beyond the pre-existing `declaration uses 'sorry'` warnings elsewhere in the project.
- [ ] xelatex blueprint check: not applicable, no `@[blueprint]` field content changes.

Made with Claude (Claude Code CLI). Tags: AI, Claude

### DIFF
--- a/PrimeNumberTheoremAnd/IwaniecKowalskiCh1.lean
+++ b/PrimeNumberTheoremAnd/IwaniecKowalskiCh1.lean
@@ -90,9 +90,11 @@ lemma unique_divisor_decomposition {a b d : ℕ} (hab : Coprime a b) (hd : d ∣
   -/)]
 theorem sum_divisors_mul_of_coprime {R : Type*} [CommRing R]
     {f : ArithmeticFunction R} (hf : f.IsMultiplicative)
-    {a b : ℕ} (hab : Coprime a b) (ha : a ≠ 0) (hb : b ≠ 0) :
+    {a b : ℕ} (hab : Coprime a b) (_ha : a ≠ 0) (_hb : b ≠ 0) :
     ∑ d ∈ (a * b).divisors, f d = (∑ d ∈ a.divisors, f d) * (∑ d ∈ b.divisors, f d) := by
-  sorry -- UPSTREAMED TO MATHLIB #36495
+  have hzf : ((zeta : ArithmeticFunction R) * f).IsMultiplicative :=
+    isMultiplicative_zeta.natCast.mul hf
+  simpa only [coe_zeta_mul_apply] using hzf.map_mul_of_coprime hab
 
 /-- If `g` is a multiplicative arithmetic function, then for any $n \neq 0$,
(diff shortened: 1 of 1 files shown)


Robby955:
closes #[1011](https://github.com/AlexKontorovich/PrimeNumberTheoremAnd/issues/1011)

Robby955:
awaiting review

Chessing234:
heads-up: the antidiagonal lemma here overlaps open #1740 / issue #1686. the other two fills look good if mathlib #36495 landed the way you use them.

Robby955:
Thanks for the comment. On mathlib #36495, this PR does not lean on it: I left
`unique_divisor_decomposition` alone because that PR covers the API, and the
three fills build against the pinned mathlib as it stands.

On the antidiagonal lemma, sure, I can drop
it from here so that #1740 carries it, leaving this PR with the other two fills and a
smaller diff to review.

Chessing234:
fyi #1740 merged the antidiagonal lemma, so dropping that hunk and rebasing should clear the conflict — the other two fills look fine.

Robby955:
I dropped that and rebased onto main, so this is now the two remaining fills.

One thing, the build of `IwaniecKowalskiCh1` is currently failing on main itself,
independently of this branch. Running the unmodified file from `d82a35e` gives the same eight
errors at the same places, none of them in the two lemmas here. They look to come from the
second, non-private `tau_sq_le_d_three` and the two `isMultiplicative_tau` identifiers
introduced in #1364. So red CI on this PR is not from this diff.

Issue #1535, opened 2026-06-12
[KADIRI]: Laplace inversion formula


teorth:
## Description

Unfortunately Laplace inversion is not currently in Mathlib.  It might be nice to put some effort into writing a nice proof here in order to upstream it to Mathlib later.

## Blueprint

[Sublemma 8.5.1](https://alexkontorovich.github.io/PrimeNumberTheoremAnd/blueprint/zeta-chapter.html#kadiri-thm-3-1-q1-laplace-inversion)

## Zulip

[Link](https://leanprover.zulipchat.com/#narrow/channel/423402-PrimeNumberTheorem.2B/topic/New.20subproject.3A.20Kadiri.27s.20explicit.20zero.20free.20region)

Robby955:
From what I understand, the proved general theorem is the principal-value/truncation form:

```lean
laplaceIntegralCpowTrunc_tendsto_of_integrable_local_quotient
```

If the exponentially weighted source is integrable and the local
quotient at `log x` is interval-integrable, then the symmetric truncated
inverse-Laplace integrals

```lean
fun T : ℝ => laplaceIntegralCpowTrunc f sigma x T
```
tend to `f (Real.log x)` as `T → ∞`.

The Kadiri-side adapter also closes in this form:

```lean
Kadiri.kadiri_thm_3_1_q1_laplace_inversion_pv
```

It uses the existing source decay plus `ContDiff ℝ 1 φ`; the local quotient
input follows from `dslope`, and no extra derivative-decay input is needed for
the PV statement.

The remaining possible mismatch is with the existing blueprinted declaration

```lean
Kadiri.kadiri_thm_3_1_q1_laplace_inversion
```

which is stated as an equality involving the whole-line Bochner integral

```lean
∫ t : ℝ, Φ (-(1+a) + t * I) * n ^ (-(1+a) + t * I)
```

That statement is reached by the local wrapper

```lean
Kadiri.kadiri_thm_3_1_q1_laplace_inversion_of_line_integrable
```

but only after adding the extra hypothesis

```lean
Integrable (fun t : ℝ =>
  laplaceIntegral φ ((-(1 + a : ℝ) : ℂ) + (t : ℂ) * I))
```

The current Kadiri hypotheses do prove the vertical-line estimate

```lean
‖laplaceIntegral φ (-(1+a) + t*I)‖ ≤ C / |t|
```

for `t ≠ 0`, via

```lean
[...]

PR #1558, opened 2026-06-13
IEANTN: bilateral Laplace inversion (#1535)

Robby955:
Converting to Draft as touches mathlib

This is an attempt at #1535 — the bilateral Laplace-inversion machinery, in a new `PrimeNumberTheoremAnd/LaplaceInversion.lean`. 

It expresses the truncated `cpow` Laplace integral as a windowed Fourier/sinc inversion, down to the von-Mangoldt-weighted limit the [Théorème](https://github.com/AlexKontorovich/PrimeNumberTheoremAnd/pull/1558) 3.1 contour estimates consume. PR awaiting.

It is a sizeable single file — the inversion passes through several intermediate kernels (Dirichlet/sinc windows, the PV split, the decay bounds) — and it currently spells the `sin(T·u)/(π·u)` kernel inline rather than through mathlib's `Real.sinc` (the value conventions at `u = 0` differ across the error- and value-form lemmas); consolidating onto `Real.sinc` is a reasonable follow-up.
This may be the similar mathlib/blueprint related issue mentioned on the Weiner Process PR.

Sorry-free; the public entry points use only the standard axioms.

Tags: AI

### DIFF
--- a/PrimeNumberTheoremAnd.lean
+++ b/PrimeNumberTheoremAnd.lean
@@ -49,6 +49,7 @@ import PrimeNumberTheoremAnd.IEANTN.ZetaAppendix
 import PrimeNumberTheoremAnd.IEANTN.ZetaDefinitions
 import PrimeNumberTheoremAnd.IEANTN.ZetaSummary
 import PrimeNumberTheoremAnd.IwaniecKowalskiCh1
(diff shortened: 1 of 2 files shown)


Robby955:
Splitting this into 5 smaller reviewable PRs; first is up as the bilateral-Laplace
  foundation.

Issue #562, opened 2026-01-12
[WIENER]: Preliminary decay bound II


teorth:
## Description

In principle a routine integration by parts; but because we are only assuming bounded variation instead of continuous differentiability, some form of Stieltjes integration may be needed.

## Blueprint

[Lemma 3.1.4](https://alexkontorovich.github.io/PrimeNumberTheoremAnd/blueprint/sect0002.html#prelim-decay-2)

## Zulip

[Link](https://leanprover.zulipchat.com/#narrow/channel/423402-PrimeNumberTheorem.2B/topic/Prime.20number.20theorem.20in.20AP/with/567400329)

teorth:
This task should wait until https://github.com/leanprover-community/mathlib-at-ICERM26/tree/stieltjes lands in Mathilb.

PR #1560, opened 2026-06-13
Wiener: preliminary decay bound II (prelim_decay_2, #562)

Robby955:
Closed: see maintainer direction 

"This particular task should wait until the Stieltjes integral theory (see https://github.com/leanprover-community/mathlib-at-ICERM26/tree/stieltjes) lands in Mathlib."

This is an attempt at #562 — it fills `prelim_decay_2` in `Wiener.lean`: the `1/|u|` decay of the Fourier transform of a bounded-variation function,

```
‖𝓕 ψ u‖ ≤ (eVariationOn ψ Set.univ).toReal / (2 * π * ‖u‖)
```

for integrable `ψ` of bounded variation and `u ≠ 0`.

The proof is the Fourier–Stieltjes integration by parts `2πiu · 𝓕ψ(u) = ∫ e(-tu) dψ(t)`, then bounding the right side by the total variation. Establishing that identity needs vector-measure-integral, bounded-variation, and Stieltjes-measure infrastructure that mathlib does not yet have, added under `PrimeNumberTheoremAnd/Mathlib/MeasureTheory/`.

It is on the larger side for that reason. The added infrastructure is general (a complex vector-measure integral, a BV→Stieltjes bridge, a finite Stieltjes IBP) and could be split into its own PR — or eventually upstreamed — if that reads better.

Sorry-free; `prelim_decay_2` uses only the standard axioms.

Developed with Claude Code CLI

### DIFF
--- a/PrimeNumberTheoremAnd.lean
+++ b/PrimeNumberTheoremAnd.lean
@@ -104,6 +104,7 @@ import PrimeNumberTheoremAnd.Mathlib.Analysis.SpecialFunctions.Pow.Deriv
 import PrimeNumberTheoremAnd.Mathlib.Analysis.SpecialFunctions.Pow.Real
 import PrimeNumberTheoremAnd.Mathlib.MeasureTheory.Integral.IntegrableOn
 import PrimeNumberTheoremAnd.Mathlib.MeasureTheory.Integral.IntervalIntegral.Basic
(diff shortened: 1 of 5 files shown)


teorth:
This particular task should wait until the Stieltjes integral theory (see https://github.com/leanprover-community/mathlib-at-ICERM26/tree/stieltjes) lands in Mathlib.  I'll add a note to the issue about this.

PR #1578, opened 2026-06-17
KADIRI: Limit of I_3(T), equation 15

Robby955:
Equation (15) of Théorème 3.1 (q = 1): the limit of I_3(T) as T tends to infinity.

The argument follows the I123 split and takes the symmetric truncated limit through the Laplace-inversion machinery now on main, with the Perron and digamma bricks (#1580, #1581) as the supporting pieces.


Robby955:
redundant, improved version coming

PR #1602, opened 2026-06-18
KADIRI: Limit of I_2(T), equation 14

Robby955:
This packages equation (14) of Théorème 3.1 (Kadiri 2005): the `T → ∞` limit of `I_2(T)`, the companion piece to equation (13).

As with (13), the argument rests on the Laplace-inversion results merged in #1579 and #1589, and the supporting lemmas are isolated in a new `KadiriEq14` module so the `Kadiri.lean` change is just the node and its wiring. This module is on the larger side; if maintainers would rather it be split further before review, let me know the suggested improvement in the code/design and I can break it up.

refactoring to v4.31.0 in progress

Tags:
AI

### DIFF
--- a/PrimeNumberTheoremAnd.lean
+++ b/PrimeNumberTheoremAnd.lean
@@ -31,6 +31,7 @@ import PrimeNumberTheoremAnd.IEANTN.FKS2
 import PrimeNumberTheoremAnd.IEANTN.FKS2Floor.Cor22Floor
 import PrimeNumberTheoremAnd.IEANTN.Goldbach
 import PrimeNumberTheoremAnd.IEANTN.Kadiri
+import PrimeNumberTheoremAnd.IEANTN.KadiriEq14
 import PrimeNumberTheoremAnd.IEANTN.KadiriSupport
 import PrimeNumberTheoremAnd.IEANTN.KLN
 import PrimeNumberTheoremAnd.IEANTN.Lcm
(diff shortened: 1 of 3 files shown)


Robby955:
awaiting-review

teorth:
I am concerned that the net size of all the additions to the Kadiri codebase may become unsustainable in terms of build times, which are again creeping up over an hour.  Is there any opportunity to reduce the net build time of these PRs by identifying common lemmas across PRs that can be reused?

Robby955:
Yes, I have some good ideas for this in progress right now.

Robby955:
I recut this on top of a small shared-support PR so the three Kadiri PRs stop re-proving each other's lemmas. KadiriEq14.lean is down from 1142 to 775 lines: the six lemmas it shared verbatim with KadiriEq13 now come from KadiriEq13 directly, the von Mangoldt tsum and summability glue and the strip bounds come from the new KadiriSupport file, and six unused imports are gone (ZetaDefinitions, KadiriZeroCounting, HadamardLogDerivative, RiemannZetaHadamard, DigammaSeries, Digamma). The statement of kadiri_thm_3_1_q1_eq_14 is unchanged.  The diff will shrink once #1619, #1620 and the support PR land.

Robby955:
I also ran mathlib's shake linter over the library: what looked to be about 20 files carrying unused imports, including PerronFormula importing Wiener; those should probably be cleaned up as well eventually, unless they are just stubs. 

EDIT: My bad, looking into this further, PerronFormula→Wiener is load-bearing, not dead (intervalIntegral_tendsto_integral comes through it)

teorth (approved):

Issue #1545, opened 2026-06-12
[KADIRI]: Limit of I_3(T) (equation (15))


teorth:
## Description

Routine consequence of the Gamma functional equation

## Blueprint

[Sublemma 8.5.11](https://alexkontorovich.github.io/PrimeNumberTheoremAnd/blueprint/zeta-chapter.html#kadiri-thm-3-1-q1-eq-15)

## Zulip

[Link](https://leanprover.zulipchat.com/#narrow/channel/423402-PrimeNumberTheorem.2B/topic/New.20subproject.3A.20Kadiri.27s.20explicit.20zero.20free.20region)

Robby955:
claim

Robby955:
propose #1604

Robby955:
The equation (15) limit merged in #1720: `kadiri_thm_3_1_q1_eq_15` in `Kadiri.lean` now closes without sorry.

PR #1604, opened 2026-06-18
KADIRI: equation 15 support modules (gamma-factor, Laplace-strip, oscillatory)

Robby955:
This is the support layer for equation (15) of Théorème 3.1 (Kadiri 2005): three small modules used by the `I_3(T)` limit, split out so the node proof that follows is a short diff.

- `KadiriEq15GammaFactor`: pole and norm bounds for the gamma factor on the relevant strip.
- `KadiriEq15LaplaceStrip`: the Laplace-strip analytic estimate.
- `KadiriEq15Oscillatory`: the oscillatory-integral helpers.

There is no change to `Kadiri.lean` here; the node that consumes these follows in a stacked PR.

Tags:
AI

Towards closure of #1545


Robby955:
awaiting-review

teorth (approved):

PR #1610, opened 2026-06-19
eSHP: correct exists_prime_gap to the Table 9 certified range (#951)

### LINKED ISSUE #951: [eSHP]: Existence of prime gaps less than 1346
## Description

Straightforward application of previous table

## Blueprint

[Proposition 10.11.10](https://alexkontorovich.github.io/PrimeNumberTheoremAnd/blueprint/secondary-chapter.html#exists-prime-gap)

## Zulip

[Link](https://leanprover.zulipchat.com/#narrow/channel/423402-PrimeNumberTheorem.2B/topic/New.20miniproject.3A.20Primes.20in.20short.20intervals)

Robby955:
This corrects the range of `exists_prime_gap` to what the eSHP Table 9 records actually certify, and proves it directly from the table.

The statement currently quantifies over `g ∈ Set.Ico 1 1476`, but Table 9's last first-gap record is at `g = 1306` (witness 3278018069102480227). That record certifies every eligible smaller gap plus 1306 itself, so the provable range is `g < 1307`; the non-record gaps from 1308 to 1344 have no certified occurrence data in the current tables. This PR sets the range to `Set.Ico 1 1307` and proves the bound from `table_9_prime_gap 1306 …`.

As with the merged Dusart `proposition_5_4b` (#1124), the proof relies on the eSHP table-certification axioms (`table_9_prime_gap`), which are the deliberate "verified by computer" inputs; it adds no new `sorry`.

If the full `< 1346` range is preferred, that needs additional certified first-occurrence data for the non-record gaps/a separate issue I think
Tags: AI

### DIFF
--- a/PrimeNumberTheoremAnd/IEANTN/eSHP/eSHP.lean
+++ b/PrimeNumberTheoremAnd/IEANTN/eSHP/eSHP.lean
@@ -454,15 +454,26 @@ theorem table_9_prime_gap_complete (g P : ℕ) (hg : g < 1346) (hg' : 0 < g)
   "exists-prime-gap"
   (title := "Existence of prime gap")
   (statement := /--
-  Every gap $g < 1346$ that is even or one occurs as a prime
+  Every gap $g < 1307$ that is even or one occurs as a prime
   gap with first prime at most $3278018069102480227$. -/)
-  (proof := /-- If not, then there would be an entry in
-  Table 8 with $P > 3278018069102480227$, which can be
-  verified not to be the case. -/)
+  (proof := /-- This follows from the Table 9 row for
+  $(1306, 3278018069102480227)$, which gives the bound for
+  every smaller even gap or the gap $1$. -/)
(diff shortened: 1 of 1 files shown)


teorth (approved):

teorth:
I assume this closes #951?

Robby955:
It's as much a fix as we can get I think at this point, rather than a full close as the issue is defined. I would say it can close the issue with misformalization related tags.

The reason is this PR, #1610, corrects exists_prime_gap to the range Table 9 certifies, the last first-gap record is g = 1306, so the provable bound is g < 1307. The full g < 1346 in #951 needs certified first-occurrence data for the non-record gaps 1308–1344, which the tables don't currently carry. To finish the task as you had described it, we would need to use Nicely's and Oliveira e Silva's tables as SHP authors only encoded up to the last record, g = 1306 (witness ~3.28×10¹⁸).

PR #1626, opened 2026-06-24

Chessing234:
## Summary
- `LSeries_totient_eq` only assumed `1 < Re(s)`, but the totient Dirichlet series converges absolutely only for `Re(s) > 2`.

## Root cause
In the strip `1 < Re(s) ≤ 2` the L-series need not converge absolutely, so the Euler product identity is false as stated.

## Fix
Strengthen the hypothesis to `2 < s.re`, matching IK (1.35).

## Test plan


Chessing234:
Superseded by #1668 (combined IK Ch1 fixes).

PR #1533, opened 2026-06-12
feat(IK): prove LSeries_liouville_eq

giuseppesorge:
## Summary
  Proves `LSeries_liouville_eq` (#1286): for `Re s > 1`, the Dirichlet series of the Liouville function `λ` equals `ζ(2s)/ζ(s)`.

  ## Approach
  `λ` is completely multiplicative with `λ(p) = -1`, so the summand `n ↦ λ(n)·n^(-s)` is a `ℕ →*₀ ℂ`. By the Euler product for completely multiplicative functions it factors as `∏ₚ (1 - λ(p)p^(-s))⁻¹ = ∏ₚ (1 + p^(-s))⁻¹`. Multiplying with the Euler product of `ζ(s)` (`HasProd.mul`) and identifying the result with that of `ζ(2s)` (`HasProd.unique`, via `(1-p^(-s))⁻¹(1+p^(-s))⁻¹ = (1-p^(-2s))⁻¹`) gives `ζ(s)·L(λ,s) = ζ(2s)`; dividing by `ζ(s) ≠ 0` yields the result. Same high-level `EulerProduct` approach as the existing `two_pow_omega` lemmas.

  ## On the helper lemmas (duplication with Mathlib)
  This file defines its own `ArithmeticFunction.liouville` rather than importing Mathlib's
  `Mathlib.NumberTheory.ArithmeticFunction.Liouville`, so Mathlib's lemmas about `liouville` are not in scope here. I
  therefore re-prove the small facts I need — `liouville_apply`, `liouville_apply_one`, `liouville_apply_mul`,
  `liouville_norm_le` — which mirror Mathlib's. The cleaner long-term fix would be to have this file use Mathlib's
  `liouville` (dropping the local definition and these helpers), but that touches the existing `liouville` definition, its
  blueprint node, and the other `liouville_*` lemmas, so I kept this PR focused on `LSeries_liouville_eq`. Happy to make that
  refactor instead if you prefer.

  ## AI assistance
  This proof was developed with AI assistance (Claude). 

Closes #1286

PR #1632, opened 2026-06-28
fix: ↗liouville in LSeries_liouville_eq and Re(s) > 1 in zeta_pow_two blueprint

Chessing234:
## Summary
- `LSeries_liouville_eq` now uses `↗liouville`, matching `↗τ`, `↗totient`, etc.
- `zeta_pow_two` blueprint states `Re(s) > 1`, consistent with the Lean hypothesis.

## Test plan
- [ ] `lake build PrimeNumberTheoremAnd/IwaniecKowalskiCh1.lean`

### DIFF
--- a/PrimeNumberTheoremAnd/IwaniecKowalskiCh1.lean
+++ b/PrimeNumberTheoremAnd/IwaniecKowalskiCh1.lean
@@ -992,7 +992,7 @@ lemma two_pow_omega_LSeries_eulerProduct_hasProd (s : ℂ) (hs : 1 < s.re) :
   "zeta_pow_two"
   (title := "zeta pow two")
   (statement := /--
-  $$\zeta(s)^2 =\zeta(2s) \sum_{n=1}^{\infty} 2^{\omega(n)} n^{-s}.$$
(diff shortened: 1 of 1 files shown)


Chessing234:
The L-series was still casting the ℤ-valued liouville function; everything else in this chapter uses ↗. Also the ζ² blueprint forgot to mention convergence.

giuseppesorge:
Heads-up on overlap with #1533: that PR proves `LSeries_liouville_eq` (currently the `sorry`), and I've just rebased it onto `main` and restated it with `↗liouville` to match the convention this PR adopts — so #1533 is now green and lands the lemma both stated *and* proved in the `↗liouville` form. This PR's change to `LSeries_liouville_eq` (statement only, keeping the `sorry`) would then be subsumed by it.

Happy to coordinate whichever order is least work for review:
- if this lands first, I'll just drop #1533's one-line statement tweak;
- if #1533 lands first, this PR's liouville hunk can be dropped (the independent `zeta_pow_two` blueprint `Re(s) > 1` fix here is unaffected either way).

No preference on order — just flagging so we don't double-edit the same line.

Chessing234:
Superseded by #1668 (combined IK Ch1 fixes).

PR #1603, opened 2026-06-18
KADIRI: recut equation (11) to the truncated-limit form (#1536)

Robby955:
Following the discussion on #1536, this recuts `kadiri_thm_3_1_q1_eq_11` to the truncated-limit form and provides the proof.

The prior whole-line Bochner-integral statement asked for an untruncated integral on the vertical line. Under the C¹ decay hypotheses, integration by parts gives only `O(1 / ‖t‖)` control for the transform, so the symmetric truncated limit is the statement those hypotheses support.

The proof is split into two small modules: `KadiriEq11Base` (the shared `I(T)` and the von Mangoldt Dirichlet-series identity) and `KadiriEq11Reduction` (the finite-window sum/integral exchange, Tannery domination, and the pointwise truncated inversion through the merged `LaplaceInversion` infrastructure). This module is also on the larger side; if maintainers would rather it be split further or if you have any shortcut or tricks to optimize the code before review, let me know and I can try to optimize/split.

Tags:
AI

Closes #1536

PR #1605, opened 2026-06-18
IEANTN: Laplace inversion cpow truncation bridge (#1535)

Robby955:
This adds the remaining cpow truncated-inversion bridge for the Laplace-inversion umbrella (#1535), after PR1-4 (#1573, #1574, #1579, #1589): `laplaceIntegralCpowTrunc_tendsto_of_integrable_local_quotient`.

It touches only `LaplaceInversion.lean` (157 added lines) and rests on the already-merged infrastructure. No other files change.

Tags:
AI

Closes #1535

PR #1634, opened 2026-07-02
KADIRI: shared support file and lemma de-privatization for the equation PRs

Robby955:
Follow-up to the build-time discussion on #1602. The three open Kadiri equation PRs re-proved lemmas that already exist in merged files or in each other, mostly because the originals were private. This PR:

- makes the six shared lemmas in KadiriEq13.lean public (no proof changes),
- makes bounded_of_continuous_of_isBigO_exp_atBot_atTop in LaplaceInversion.lean public,
- adds IEANTN/KadiriSupport.lean: signed-form corollaries of the KadiriEq12Foundations full-strip lemmas, the shared weighted-source bounds, the von Mangoldt summability glue, and a convex-set form of the sinc window bound,
- moves tsum_vonMangoldt_eq from Kadiri.lean into KadiriSupport and removes three private duplicates of the Eq13 lemmas that Kadiri.lean itself carried,
- repairs five proofs in KadiriEq12Foundations.lean (statements unchanged). That module is currently outside the root import closure: nothing imports it, so CI never elaborates it, and it broke silently in the June 20 bump to v4.31.0 / mathlib db127794. This PR repairs the proofs and, via KadiriSupport's import, brings the module back into the build.

Stacked on #1619 and #1620, so the diff shrinks when those merge. With this in place the recuts of #1602, #1603 and #1604 drop about 1100 lines total, and each shared proof elaborates once instead of two or three times.


Robby955:
While reviewing this PR, I thought I should include this while I verify it, I found that KadiriEq12Foundations.lean is not in the root import closure, so CI has never elaborated it. It broke in the June 20 toolchain bump and is currently red on main if built directly; this PR repairs it and pulls it into the build. Five other files are also outside the closure (CriticalLineDecay and the four Unused files, possibly intentional, could be stubs). A CI step like mathlib's lake exe mk_all check could catch this class of silent breakage.

teorth:
Thanks!  This is a significant cleanup.

What order of merging would you recommend for the PRs, and do any past PRs now need to be closed for this refactoring?

Robby955:
For Order:

 #1619 then #1620, then #1634.

Then #1602, #1603, #1604 last their recut commits are already in with green CI, so that step is just rebase and force-push once #1634 lands, not new proof work.

This repo squash-merges everything, so none of that shrinkage happens automatically; I'll rebase each as the previous piece lands. 

Nothing needs closing. #1601 and #1605 are already merged, the rest will get rebased, not reopened. 

#1634's build is green now, so it's ready whenever you want to merge it as well.

teorth (approved):

PR #1647, opened 2026-07-18
fix: FKS (3.13) gamma range is H_a ≤ γ < T

Chessing234:
## Summary
- `eq_13`: `Set.Ioc` → `Set.Ico` so the sum is over $H_a \le \gamma < T$ as in FKS (3.13).
- Prop 3.6 blueprint: $x^{\sigma_1-1}$, not $x_1^{\sigma_1-1}$.

## Test plan
- [ ] Check against arXiv:2204.02588 (3.13) and (3.8)


Chessing234:
awaiting-review

Chessing234:
Superseded by #1671 (combined FKS fixes).

PR #1648, opened 2026-07-18
fix: classical ZFR, RvM, and ZDB blueprint statements

Chessing234:
## Summary
- Classical zero-free region: $\Re(s)\ge 1-1/(R\log|t|)$, matching Lean.
- Riemann–von Mangoldt: restore the absolute value around the main term.
- Zero-density bound: drop the pasted RvM fragment; state the $N(\sigma,T)$ estimate.

## Test plan
- [ ] Diff blueprints against the Lean defs in `ZetaDefinitions.lean`


Chessing234:
awaiting-review

Chessing234:
Superseded by #1669 (combined Defs/Zeta fixes).

PR #1653, opened 2026-07-18
fix: JY Thm 1.4 uses Eπ exponent 1.801

Chessing234:
## Summary
- Paper gives $|\pi-\mathrm{Li}|\le 0.028\,x\,(\log x)^{0.801}\cdots$; `Eπ` divides by $x/\log x$, so the Vinogradov exponent is $1.801$.
- Blueprint range $x\ge 23$ to match the Lean statement / paper.

## Test plan
- [ ] Check against arXiv:2204.01980 Thm 1.4 and `Eπ` in `Defs.lean`


Chessing234:
Superseded by #1673 (combined secondary-estimate fixes).

PR #1663, opened 2026-07-18
fix: PT/JY π bounds use li

Chessing234:
## Summary
- PT Cor 2 and JY Cor 1.3 blueprints: $|\pi-\mathrm{li}|$ as in the papers.

## Test plan
- [ ] Check arXiv:1809.03134 Cor 2 and arXiv:2204.01980 Cor 1.3

Made with [Cursor](https://cursor.com)

PR #1661, opened 2026-07-18
fix: JY ψ bound exponent is 1.515

Chessing234:
## Summary
- Use $(\log x)^{1.515}$ as in JY Thm 1.1 / Table 1 (was $1.51$).

## Test plan
- [ ] Check arXiv:2204.01980 Thm 1.1

PR #1659, opened 2026-07-18
fix: PT ψ bound uses Table 1 constants

Chessing234:
## Summary
- Replaced the Corollary 2 $\pi$ constants ($235$, $0.52$) with Theorem 1 / Table 1 for $\psi$ at $e^{2000}$.

## Test plan
- [ ] Check arXiv:1809.03134 Thm 1, Table 1, Cor 2

PR #1652, opened 2026-07-18
fix: Trudgian Thm 2 uses |π − li|

Chessing234:
## Summary
- State $|\pi(x)-\mathrm{li}(x)|$ as in Trudgian Thm 2.
- Drop the `Eπ`/`Li` encoding (wrong comparison function and scaling).

## Test plan
- [ ] Compare to arXiv:1401.2689 Thm 2

Made with [Cursor](https://cursor.com)

PR #1673, opened 2026-07-18
fix: PT/JY/Trudgian secondary estimate statements

Chessing234:
## Summary
- Trudgian Thm 2: `|π-li|`
- JY Thm 1.4: `Eπ` exponent `1.801`; PT/JY π blueprints use `li`
- PT ψ Table 1 constants; JY ψ exponent `1.515`

Supersedes #1652 #1653 #1659 #1661 #1663.

## Test plan
- [ ] Compare PT / JY / Trudgian papers for the cited theorems

### DIFF
--- a/PrimeNumberTheoremAnd/IEANTN/SecondarySummary.lean
+++ b/PrimeNumberTheoremAnd/IEANTN/SecondarySummary.lean
@@ -130,7 +130,7 @@ theorem corollary_1 (X σ A B C ε₀ : ℝ) (h : (X, σ, A, B, C, ε₀) ∈ Ta
   (statement := /--
   One has
   \[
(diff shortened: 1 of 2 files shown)


teorth (approved):

teorth:
The Johnston-Yang typo is also reproduced at https://archimede.pages.math.cnrs.fr/tme-emt-wiki/Art01.html#bounds-on-primes-in-special-ranges .  If your agent is able to compile a list of such errata into a single PDF document, this might be worth submitting to the TME-EMT moderators at tmeemt.wiki@gmail.com . (But, as with the PRs here, I strongly recommend using a single document rather than sending dozens of small updates.)

PR #1714, opened 2026-07-25
Chore (MediumPNT): some proof simplification

ajirving:
Drops I6-I9 bounds in favour of immediately rewriting using conjugation.  That decreases the number of constants we need to use in the MediumPNT proof.  I also removed the increase to max heartbeats since that doesn't seem to be needed.


Chessing234:
nice cleanup — folding i6–i9 into the conj symmetries and dropping the heartbeat bump looks right. only ask is whether the blueprint nodes for those dropped bound lemmas need a follow-up so the informal side stays in sync.

ajirving:
I can't see any mentions of those dropped results in the blueprint which need to be updated.  They were only used in the proof of MediumPNT for which the blueprint proof just says "evaluate the integrals".

Issue #1257, opened 2026-04-05
[BKLNW]: Verify Table 11


teorth:
## Description

A lengthy numerical computation

## Blueprint

[Proposition 10.8.5](https://alexkontorovich.github.io/PrimeNumberTheoremAnd/blueprint/secondary-chapter.html#bklnw-table-11-verification)

## Zulip

[Link](https://leanprover.zulipchat.com/#narrow/channel/423402-PrimeNumberTheorem.2B/topic/Starting.20on.20BKLNW)

Robby955:
For Table 11, I think the encoded target and the printed values may come from different parts of the paper.

The repo's table_11 values for k = 1..3 at small b₀ match the C_{b,k} table from Corollary 9.1 in the full-tables ancillary file (valid for e^b ≤ x ≤ 10^19, computed with c = −0.8, C = 0.81), while the k = 4, 5 values match the cumulative strip maxima. As encoded, B_8_1' takes the max of B_8_1 over the Table 10 strips only, and already at b = 20 I get B_8_1 for k = 1 around 1.85e-3, while the listed value is 1.6844e-3; the exact interval supremum gives about 1.81e-3, so it does not close the gap either.

So I think the printed values are the max of the Corollary 9.1 bound below 10^19 and the strip bound above it, which the current strip-only encoding cannot reproduce for small b₀. I may be misreading the intended definition, but this seems to be the discrepancy I'm hitting.

giuseppesorge:
Following up on @Robby955's observation above (2026-06-10): I looked into `bklnw_table_11_verification` and I think the diagnosis is right — the theorem as currently stated isn't provable against the present `table_11` data.

The encoded target compares the printed `table_11` value `B k` against `B_8_1' k b₀` (`BKLNW.lean:1270`), i.e. the supremum of `B_8_1` over the Table 10 strips **only**. But the `table_11` entries for `k = 1..3` at small `b₀` are the Corollary 9.1 constants `C_{b,k}` (valid for `e^b ≤ x ≤ 10^19`), not strip maxima. Concretely, as Robby955 noted, at `b₀ = 20, k = 1` the strip supremum is ≈ 1.81e-3, which already exceeds the listed `1.6844e-3`, so `B_8_1' k b₀ ≤ B k` is false for that row.

[...]

PR #1727, opened 2026-07-31
feat(BKLNW): prove bklnw_table_11_verification as a split-domain θ-bound (#1257)

giuseppesorge:
Closes the `sorry` on `bklnw_table_11_verification` (#1257).

## Why the statement changed

The encoded target was **false**, as @Robby955 diagnosed on the issue (2026-06-10). It compared the printed Table-11 values against `B_8_1' k b₀` — the Table-10 strip supremum taken from `b₀` itself. But a Table-11 entry is a maximum over *two* regimes split at `10^19`: below it the entries are the Corollary-9.1 constants of Table 12, above it the Table-10 strip supremum. Taking the strip supremum from `b₀` sweeps in strips below `10^19`, where the smaller Corollary-9.1 constant is the one that applies. At `b₀ = 20, k = 1` that supremum is `≈ 1.81e-3`, above the printed `1.6844e-3`.

@teorth greenlit restating it on 2026-07-07 ("Sounds good to me, please go ahead and restate Table 11"). The new statement is the split-domain θ-bound the table actually asserts:

```lean
theorem bklnw_table_11_verification (b₀ : ℝ) (B : ℕ → ℝ)
    (h : (b₀, B 1, B 2, B 3, B 4, B 5) ∈ BKLNW.table_11) :
    ∀ k ∈ Finset.Icc 1 5, ∀ x ∈ Set.Icc (exp b₀) (exp K),
      |θ x - x| ≤ B k * table_11_margin * x / (log x) ^ k
```

No added hypotheses. There were no consumers of the old statement.

**Correction to my own status comment.** My 2026-07-27 comment on #1257 described the above-`10^19` route as going through "the strip supremum `B_8_1'`". That is wrong and I never acted on it: `B_8_1'` is the Corollary-8.1 *endpoint surrogate*, which the printed Table-10 values are provably not upper bounds for (#1255). The route used here goes through `B_8_exact` and the proven `bklnw_table_10_verification`.

## Why a margin factor

As with Table 10, the printed values hold only up to rounding, so the bound carries `table_11_margin`. Sweeping all 43 rows × 5 columns, the tightest obligations are:

- **above `10^19`**: `b₀ = 20, k = 4`, where the Table-10 suffix supremum and the Table-11 entry are the *same* printed number `5.7184e1`, so the requirement is exactly `table_10_margin ≤ table_11_margin`;
- **below `10^19`**: `b₀ = 25, k = 1`, needing `≥ 1.000107` (Table 12's row 25 was itself corrected upward from the paper — see the correction note on `table_12`).

**So `table_11_margin = table_10_margin` would already suffice.** I kept the chained value `table_10_margin * 1.001 = 1.003003001` to follow the per-table convention (`table_10_margin = table_8_margin * 1.001`) and leave headroom — the tight choice makes the `b₀ = 20, k = 4` obligation an exact equality, which breaks on any re-rounding of a table entry. The cost is a constant 0.1% weaker than strictly necessary. Happy to tighten it if you prefer the sharper constant.

## Why the theorem moved file

It could not be proved where it was stated. Its above-`10^19` branch needs `bklnw_table_10_verification`, which lives in `BKLNW_table10_dispatch.lean`; but the Table-10 row files import `BKLNW.lean` (for `B_8_exact`), so consuming it from inside `BKLNW.lean` is circular. It now lives in a new `BKLNW_table11_dispatch.lean`, downstream of the Table-10 dispatch — exactly where `bklnw_table_10_verification` itself lives, for the same reason. `BKLNW.lean` keeps a pointer comment.

## How it is proved

**Above `10^19`.** `theta_bound_of_table_10_suffix` is the `B_8_exact` analogue of `bklnw_cor_8_1b`: the same coverage argument, routed through the exact Lemma-8 interval supremum instead of the endpoint surrogate. It reduces the whole range to a rational obligation over a *suffix* of Table 10.

`10^19 = e^{19 log 10}` is itself a Table-10 grid node (row 24, between rows `43` and `44`), so the split lands on a strip boundary and no partial-strip handling is needed.

The suffix obligation would naively pit each of the 43 Table-11 rows against each of the 287 Table-10 rows. Instead `eq_or_ge_next` — no Table-10 entry lies strictly between `b` and `table_10_next b`, immediate from `table_10_next` being an infimum — turns it into a single descending walk of the grid: `table_11_suffix_from_<b>` records the running suffix maximum and follows from its own successor in one step. 263 cheap steps instead of ~12000 comparisons; the chain elaborates in ~11s.

**Below `10^19`.** Corollary 9.1 is one-sided — it bounds `x - θ x` and says nothing about `θ x - x` — so it cannot on its own bound `|θ x - x|`. The missing half is `bklnw_eq_3_17` (`θ x < x - 0.05 √x` for `x ≤ 10^19`): below `10^19` the absolute value collapses to `x - θ x`, which is exactly Corollary 9.1's subject. Of the 43 rows, 12 are live (`b₀ = 20..30` and `43`, each with a Table-12 row); the other 31 start above the node and are vacuous.

**The `b₀ = K` row** has the degenerate domain `[e^K, e^K]` and is not itself a Table-10 entry (Table 10 stops at `24000`). It is served by the last real strip `[24000, K]` through the same engine.

The generated chain and its per-row certificates live in their own file, following the `BKLNW_table10_rows_*` convention; the dispatch file keeps only the hand-written argument.

## Axiom surface

Both new files are `sorry`-free. The 263-lemma chain together with its dispatch (`table_11_suffix_dominates`, `table_11_top_row_dominates`) needs exactly `propext` / `Classical.choice` / `Quot.sound` plus **two** `native_decide` axioms, and those two are inherited from the pre-existing `LogTables.log_10_gt` / `log_10_lt`: placing the `10^19` node `19 log 10` on the Table-10 grid needs numeric bounds on `log 10`. This PR introduces **no new** `native_decide` obligation, and the grid walk itself is clean — `eq_or_ge_next` needs none, since it only compares printed values against printed values.

*(Correction: an earlier revision of this section claimed the chain needed "not even `native_decide`". That was wrong — the two `LogTables` axioms above were always there.)*

The theorem as a whole still reports `sorryAx`, but **entirely from pre-existing upstream stubs, on both branches**: `cor_5_1` (via `bklnw_cor_8_1a_exact`) above `10^19`, and `Buthe.theorem_2a` / `2c` (via `bklnw_eq_3_17`) below it. `bklnw_table_10_verification` already carried the same contingency on `main`. This branch adds no `sorry` of its own.

## Test plan

- [x] `lake build` of the whole project: **4342 jobs, 0 errors**, no warnings from either new file.
- [x] `#print axioms BKLNW.bklnw_table_11_verification` — inspected; see above.
- [x] Coverage: the theorem instantiated at **all 43 rows × 5 columns = 215/215**, each elaborating to a concrete, hypothesis-free bound. This rules out a vacuous hypothesis or a silently uncovered row.
- [x] The numeric content was independently pre-computed outside Lean over the parsed tables before generating any proof, and agrees with what Lean checks.
- [x] Rebased on current `main` (`21998bb`, Lean v4.32.2, Mathlib v4.32.2, LeanCert v4.32.2.2).

## Notes

Credit to @Robby955 for the original diagnosis that the encoded statement was false.

Made with Claude Code.

### DIFF
--- a/PrimeNumberTheoremAnd.lean
+++ b/PrimeNumberTheoremAnd.lean
@@ -17,6 +17,7 @@ import PrimeNumberTheoremAnd.IEANTN.BKLNW.BKLNW_app
 import PrimeNumberTheoremAnd.IEANTN.BKLNW.BKLNW_app_tables
 import PrimeNumberTheoremAnd.IEANTN.BKLNW.BKLNW_table10_rows
 import PrimeNumberTheoremAnd.IEANTN.BKLNW.BKLNW_table10_dispatch
+import PrimeNumberTheoremAnd.IEANTN.BKLNW.BKLNW_table11_dispatch
 import PrimeNumberTheoremAnd.IEANTN.BKLNW.BKLNW_tables
 import PrimeNumberTheoremAnd.IEANTN.Buthe
 import PrimeNumberTheoremAnd.IEANTN.CH2.CH2
--- a/PrimeNumberTheoremAnd.lean
+++ b/PrimeNumberTheoremAnd.lean
@@ -18,6 +18,7 @@ import PrimeNumberTheoremAnd.IEANTN.BKLNW.BKLNW_app_tables
 import PrimeNumberTheoremAnd.IEANTN.BKLNW.BKLNW_table10_rows
 import PrimeNumberTheoremAnd.IEANTN.BKLNW.BKLNW_table10_dispatch
 import PrimeNumberTheoremAnd.IEANTN.BKLNW.BKLNW_table11_dispatch
+import PrimeNumberTheoremAnd.IEANTN.BKLNW.BKLNW_table11_suffix
 import PrimeNumberTheoremAnd.IEANTN.BKLNW.BKLNW_tables
 import PrimeNumberTheoremAnd.IEANTN.Buthe
 import PrimeNumberTheoremAnd.IEANTN.CH2.CH2
--- a/PrimeNumberTheoremAnd/IEANTN/BKLNW/BKLNW.lean
+++ b/PrimeNumberTheoremAnd/IEANTN/BKLNW/BKLNW.lean
@@ -1509,40 +1509,12 @@ theorem bklnw_cor_8_1b (k : ℕ) (b₀ : ℝ) (hk : 1 ≤ k ∧ k ≤ 5)
       (mul_le_mul_of_nonneg_right hB hx_pos.le)
       (pow_nonneg (log_pos hx_gt_one).le k)
 
-@[blueprint
-  "bklnw-table-11-verification"
-  (title := "BKLNW Table 11 verification")
-  (statement := /--  Verification of the entries of Table 11. -/)
-  (proof := /-- TODO: Implement a margin and verify the entries of Table 11. Any lengthy numerical calculations should be moved to `BKLNW\_tables.lean`-/)
-  (latexEnv := "proposition")
-  (discussion := 1257)]
-theorem bklnw_table_11_verification (b₀ : ℝ) (B : ℕ → ℝ)
-    (h : (b₀, B 1, B 2, B 3, B 4, B 5) ∈ BKLNW.table_11) :
-    ∀ k ∈ Finset.Icc 1 5, ∀ x ∈ Set.Icc (exp b₀) (exp K),
-      |θ x - x| ≤ B k * table_11_margin * x / (log x) ^ k := by
-  intro k hk x hx
-  by_cases hsplit : x ≤ (10 : ℝ) ^ 19
-  · -- Below 10^19: the Table-11 entries in this regime are the Corollary-9.1 constants,
-    -- so the bound comes from `bklnw_corollary_9_1` / `bklnw_corollary_9_1_explicit`,
-    -- which give `|θ x - x| ≤ C_bk b₀ c C c₀ k · x/(log x)^k` on `[exp b₀, v]` for a
-    -- Butzer row `(100, v, c, C) ∈ table_from_buthe` with `v ≥ 10^19`. The residual
-    -- per-row obligation is the all-rational `C_bk b₀ c C RS_prime.c₀ k ≤ B k * table_11_margin`,
-    -- of exactly the shape already discharged by `bklnw_table_12_verification`.
-    -- Note this branch is vacuous for rows with `b₀ ≥ 19 * log 10`.
-    sorry
-  · -- Above 10^19: route through the *exact* Lemma-8 supremum, not the endpoint surrogate
-    -- `B_8_1'` — Table 10's printed values are provably exceeded by `B_8_1` (issue #1255),
-    -- so only `B_8_exact` chains into the tables. For `x > 10^19` we have
-    -- `log x ≥ max b₀ (19 * log 10)`, so `table_10_coverage` picks a Table-10 strip
-    -- `b ≥ max b₀ 43` with `x ∈ [exp b, exp (table_10_next b)]`; then
-    -- `bklnw_cor_8_1a_exact k b (table_10_next b)` bounds `|θ x - x|` by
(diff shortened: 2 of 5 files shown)


Chessing234:
big but the false-statement → split-domain rephrase matches the #1257 diagnosis. didn't recheck the table certificates line-by-line; main ask is whether the dispatch/suffix split is the intended long-term layout or a temporary scaffolding dump.

giuseppesorge:
Thanks for looking. Short answer: intended long-term layout — it's the shape the repo documents for exactly this case, not scaffolding.

**`PULL_REQUEST_STYLE.md` §5 names this architecture as the worked example:**

> **Very large certificate corpora get sharded across many small files** — the BKLNW Table 10 architecture (~22 files `BKLNW_table10_rows_*` + `_dispatch` + `_next` + `_values`) is the worked example. Sharding lets Lake compile shards in parallel and lets a downstream edit invalidate only the shard(s) it touches, not the whole table.

> **Don't inline a 100-row certificate table into a proof file** because it's "simpler". It isn't — the next contributor pays the compile cost of your convenience every time they touch anything in the paper file.

So `BKLNW_table11_dispatch.lean` (526 lines: statement + hand-written dispatch) over `BKLNW_table11_suffix.lean` (generated certificates) is the same two-layer shape as the merged `BKLNW_table10_dispatch.lean` (4338 lines, header: *"Generated proof of the Table 10 verification theorem from proved row/k margins"*) over its `_rows_*` files. Folding it back into `BKLNW.lean` is what that second bullet warns against — and it can't go there anyway: the above-`10^19` branch consumes `bklnw_table_10_verification`, while the Table-10 row files import `BKLNW.lean`, so it would be circular.

**Why 263 lemmas rather than one induction.** They aren't copies — each `table_11_suffix_from_<b>` carries a different constant vector (the running suffix maximum of the five printed columns from that node on) and is proved from its own successor in one step. An induction wouldn't collapse it either: the grid successor is encoded upstream as 287 separate `table_10_next_cert_<b>` lemmas in `BKLNW_table10_next.lean`, so it would still consume them node by node. Unifying that means re-encoding Table 10 itself — merged code, out of scope here. Happy to take it as a follow-up if it's wanted; it would clean up Table 10 and Table 11 together.

**One place I did deviate from §5, deliberately.** The guide says "many small files"; the suffix chain is one file. The sharding rationale doesn't bite here: the chain is strictly sequential (each node depends on its successor), so there is no Lake parallelism to win and no useful incremental invalidation — touching any node invalidates everything below it regardless. It elaborates in 11s, against the heavy interval arithmetic in the Table-10 row files. If you'd rather it match Table-10 granularity, say so and I'll shard it.

I've recorded the above in the dispatch file header so the next reader doesn't have to ask.

---

**Separately — a correction to my own PR body, which you shouldn't have to catch.**

The Axiom surface section claimed the chain needs *"not even `native_decide`"*. That is wrong. `table_11_suffix_dominates` and `table_11_top_row_dominates` inherit two `native_decide` axioms from `LogTables.log_10_gt` / `log_10_lt`, because placing the `10^19` node `19 log 10` on the Table-10 grid needs numeric bounds on `log 10`. Both are pre-existing shared lemmas and this PR introduces no new `native_decide` obligation — the grid walk itself is clean (`eq_or_ge_next` needs none). Corrected in the file header and in the PR body above.

Also rebased onto current `main`: the toolchain moved to v4.32.2 / LeanCert v4.32.2.2 since I opened this. Full build green — **4342 jobs, 0 errors**, no warnings from either new file. (The bump's new `linter.style.maxHeartbeats` wants the justifying comment *below* the `set_option` rather than above it; the comment was already there, just in the wrong place. Moved.)

On not rechecking the certificates line-by-line — entirely reasonable, and the chain is the cheap part to trust: it only ever compares printed values against printed values.

teorth (approved):

